PR #11, opened 2024-06-06
Support recursive circuit format, single witness loading; support const in verification

sixbigsquare:


src/exec.rs
```diff
@@ -0,0 +1,87 @@
...
+        );
+        return;
+    }
+    let command = &args[1];
+    let circuit_file = &args[2];
+    let witness_file = &args[3];
+    let output_file = &args[4];
+    match command.as_str() {
+        "prove" => {
+            let config = Config::new();
```
zhenfeizhang, line 63: CI failed because the api has changed here. see https://github.com/PolyhedraZK/Expander-rs/blob/9f0fbdc1c1e372376b8c2a5af66684f16ffbc791/src/main.rs#L43

niconiconi:
CI doesn't pass

PR #23, opened 2024-07-08
BN254 CLI support

sixbigsquare:
Note: it seems that some testing files may have been outdated (which may affect testing):

https://github.com/PolyhedraZK/Expander-rs/blob/381fa8041be30bc0b3df9abf4863b4bf161fd3f4/readme.md?plain=1#L23-L24


arith/src/field/m31/vectorized_m31.rs
```diff
@@ -47,6 +47,15 @@ impl FieldSerde for VectorizedM31 {
             }
         }
     }
+
+    #[inline(always)]
+    fn deserialize_from_ecc_format(bytes: &[u8; 32]) -> Self {
+        let val: [u8; 32] = bytes[0..32].try_into().unwrap();
```
zhenfeizhang, line 53: seems `val` is not needed

``` rust
        for (i, v) in bytes.iter().enumerate().skip(4).take(28) {
            assert_eq!(*v, 0, "non-zero byte found in witness at {}'th byte", i);
        }

        Self::from(u32::from_le_bytes(bytes[..4].try_into().unwrap()))
```

PR #38, opened 2024-07-25
Impl config trait for better abstraction

zhenfeizhang:


src/circuit.rs
```diff
@@ -271,7 +271,7 @@ impl<F: Field + SimdField> Segment<F> {
         for _ in 0..rand_coef_idx_num {
             let idx = u64::deserialize_from(&mut reader) as usize;
 
-            let rand_coef = t.challenge_f::<F>();
+            let rand_coef = t.challenge_f::<C>().first_base_field(); // ZZ: THIS SEEMS SUSPICIOUS
```
zhenfeizhang, line 274: Is this right? So the circuit rand has less entropy? hmmm maybe we need three of those?

PR #49, opened 2024-08-11
fix inv of 2

zhenfeizhang:
- use a function call instead of a constant
- default implementation is unimplemented so `F_{2^m}` doesn't need to implement `inv_of_2`


zhenfeizhang:
not required

PR #60, opened 2024-08-17
Minor: Try to resolve some TODOs in neon

tonyfloatersu:


arith/src/extension_field/gf2_128/neon.rs
```diff
@@ -176,27 +176,26 @@ impl BinomialExtensionField for NeonGF2_128 {
...
-        // res
+    fn add_by_base_field(&self, base: &Self::BaseField) -> Self {
+        if base.is_zero() {
+            return *self;
+        }
+        add_internal(&Self::one(), self)
     }
 
     #[inline(always)]
     fn first_base_field(&self) -> Self::BaseField {
```
niconiconi, line 195: We should remove this function

niconiconi (changes requested):

arith/src/extension_field/gf2_128/neon.rs
tonyfloatersu: wdym? take away the trait method together with all the implementations for `first_base_field`?

PR #62, opened 2024-08-22
Minor: CI process simplify with `dev-setup` loading circuits

tonyfloatersu:
Gather all the scattering definitions to local circuits, and load circuits from remote google storage.


niconiconi:
don't use sudo in CI

PR #64, opened 2024-08-26
Minor: Serialization and Deserialization Error Propagation

tonyfloatersu:
Minor PR, containing following changes:
- Exposing `FieldSerdeError` from `std::io::Error` with `thiserror`, serialize and deserialize return with `Result` rather than `unwrap`.
- Moving `serialized_size` trait method into a trait constant `SERIALIZED_SIZE: usize`.  The thinking here is, the type implementing `FieldSerde` has a const size that is known in compile time, so the size should be known from the type rather than an instance.

### DIFF
--- a/src/circuit.rs
+++ b/src/circuit.rs
@@ -255,9 +255,9 @@ impl<C: GKRConfig> Segment<C> {
             || !self.gate_uni.is_empty()
     }
 
-    pub(crate) fn read<R: Read>(mut reader: R) -> Self {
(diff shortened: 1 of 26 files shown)


src/circuit.rs
```diff
@@ -123,12 +123,11 @@ impl<C: GKRConfig> Clone for Circuit<C> {
     }
 }
 
-// FIXME: not 100% sure this is correct
```
tonyfloatersu, line 126: nit: I think this should be right, just checked a few docs from rust official

PR #78, opened 2024-08-28
Gf2 dev

zhiyong1997:


zhiyong1997:
might need to revert 'revert mul by x' in the 'main' branch for fast forward.

zhiyong1997:
It goes up from 8.7k/s to 9k/s with`mul-by-x`. I would prefer to merge this into the main branch first, and tune the performance later (to avoid additional merging conflict.)

PR #91, opened 2024-09-12
impl mul by i32

zhenfeizhang:


zhenfeizhang:
this method is not faster

PR #108, opened 2024-09-26
Public input rebase

zhiyong1997:


zhiyong1997:
We need a poseidon circuit in the new serialization format in order to pass the CI.

src/circuit/expander_circuit.rs
```diff
@@ -1,4 +1,8 @@
-use arith::Field;
+use core::panic;
```
zhenfeizhang: nits: don't need to `use core::panic` -- its imported by default

src/circuit/serde.rs
```diff
@@ -0,0 +1,216 @@
...
+    fn deserialize_from<R: Read>(mut reader: R) -> Self {
+        let gate_type = <usize as FieldSerde>::deserialize_from(&mut reader).unwrap();
+        let i_ids: [usize; INPUT_NUM] = Vec::<usize>::deserialize_from(&mut reader)
+            .try_into()
+            .unwrap();
+
+        let o_id = <usize as FieldSerde>::deserialize_from(&mut reader).unwrap();
+
+        let coef_type_u8 = u8::deserialize_from(&mut reader).unwrap();
+        let (coef_type, coef) = match coef_type_u8 {
```
zhenfeizhang, line 101: does this start from `0` or `1`?

zhiyong1997: It starts from 1 because that is how it's defined in ECC. We can switch to 0 though.

benches/gkr_hashes.rs
```diff
@@ -66,9 +84,9 @@ fn criterion_gkr_keccak(c: &mut Criterion) {
 
 fn criterion_gkr_poseidon(c: &mut Criterion) {
```
zhenfeizhang, line 85: https://github.com/PolyhedraZK/Expander/actions/runs/11076085794/job/30933755923

looks like rust ECC doesn't generate Poseidon circuit. The old Poseidon circuit triggered failure

zhenfeizhang (approved):

PR #122, opened 2024-10-14
Fix expander-exec

mcfx:
Tested `prove` and `verify` with `gkr/data/circuit-*.txt`, all 3 fields work well.

Also tested `serve` with current compiler ([here](https://github.com/PolyhedraZK/ExpanderCompilerCollection/blob/master/ecgo/integration/expander_exec.go#L49)), it works.


zhenfeizhang (approved):
LGTM

PR #133, opened 2024-10-28
building blocks for basefold pcs

zhenfeizhang:
this PR adds required building blocks for basefold. include the following:

- wrapper for plonky3's babybear -- this may needs further refactoring for our SIMD interfaces
- implemented poseidon hash function for both babybear and mersenne
- add classic sumcheck for arbitrary gate
- add an implementation of merkle tree
- refactoring APIs for univariate polynomials


arith/babybear/src/baby_bear.rs
```diff
@@ -0,0 +1,229 @@
...
+        Self::new((x % Self::modulus()).as_u32())
+    }
+
+    fn to_u256(&self) -> ethnum::U256 {
+        // Converts to canonical form before casting to U256
+        ethnum::U256::from(self.as_u32_unchecked())
+    }
+}
+
+// TODO: Actual SIMD impl
```
tonyfloatersu: nit: are we good to remove this TODO, as we already have avx/neon impls?

arith/babybear/src/baby_bear.rs
```diff
@@ -0,0 +1,229 @@
...
+
+    const SIZE: usize = 32 / 8;
+
+    const FIELD_SIZE: usize = 32;
+
+    const ZERO: Self = BabyBear::new(0);
+
+    const ONE: Self = BabyBear::new(1);
+
+    // See test below
```
tonyfloatersu: nit: remove - tests moved to a single file

.github/workflows/ci.yml
```diff
@@ -18,6 +18,8 @@ jobs:
       - run: python3 ./scripts/install.py
       - run: cargo fmt --all -- --check
       - run: cargo clippy --all
+        env:
+          RUSTFLAGS: "-Dwarnings -C target-cpu=native -C target-feature=+avx512f"
```
tonyfloatersu, line 23: wait... do we need these options for linting?

arith/polynomials/src/univariate/coeff.rs
```diff
@@ -0,0 +1,111 @@
...
+        let mut res = self.coefficients.clone();
+        let omega = primitive_root_of_unity(self.coefficients.len());
+        let log_n = self.coefficients.len().trailing_zeros();
+        best_fft(&mut res, omega, log_n);
+        res
+    }
+}
+
+/// Compute poly / (x-point) using univariate division
+pub fn univariate_quotient<F: Field>(poly: &[F], point: &F) -> Vec<F> {
```
tonyfloatersu, line 82: note taking: this is from original `bi-kzg/src/poly/utils.rs`

arith/polynomials/src/bivariate/lagrange.rs
```diff
@@ -1,10 +1,34 @@
...
 
-use super::{lagrange_coefficients, BivariateLagrangePolynomial, BivariatePolynomial};
-use halo2curves::ff::{Field, PrimeField};
-use rand::RngCore;
+use super::BivariateLagrangePolynomial;
+
+/// For a point x, compute the coefficients of Lagrange polynomial L_{i}(x) at x, given the roots.
+/// `L_{i}(x) = \prod_{j \neq i} \frac{x - r_j}{r_i - r_j}`
+#[inline]
+pub fn lagrange_coefficients<F: FFTField + Send + Sync>(roots: &[F], points: &F) -> Vec<F> {
```
tonyfloatersu: note taking: this is from original `bi-kzg/src/poly/utils.rs`

arith/polynomials/src/bivariate.rs
```diff
@@ -0,0 +1,38 @@
...
+    }
+}
+
+impl<F: FFTField> From<&BivariatePolynomial<F>> for BivariateLagrangePolynomial<F> {
+    #[inline]
+    fn from(poly: &BivariatePolynomial<F>) -> Self {
+        let coeffs = poly.interpolate();
+        BivariateLagrangePolynomial::new(coeffs, poly.degree_0, poly.degree_1)
+    }
+}
```
tonyfloatersu, line 39: note taking: this chunk is originally from `arith/polynomials/src/bivariate/lagrange.rs`

arith/polynomials/src/bivariate/lagrange.rs
tonyfloatersu: nits:
1. I suppose this is actually a univariate case, so this method fits better under `univariate/lagrange.rs` or `utils.rs`?
2. suggest renaming `points` to `point`, as you are evaluating at a single point for univariate lagrange basis determined by `roots`.

arith/polynomials/src/bivariate/lagrange.rs
```diff
@@ -1,10 +1,34 @@
...
+        .map(|(i, _)| {
+            let mut numerator = F::ONE;
+            let mut denominator = F::ONE;
+            for j in 0..roots.len() {
+                if i == j {
+                    continue;
+                }
+                numerator *= roots[j] - points;
+                denominator *= roots[j] - roots[i];
+            }
```
tonyfloatersu: is this right...? why I think this should be something like `numerator *= (roots[j] - points)` with bracket on?

arith/polynomials/src/bivariate/lagrange.rs
tonyfloatersu: nit: please also specify this is a brute-force version of implementation, that has O(n^2) runtime complexity

arith/polynomials/src/bivariate/lagrange.rs
tonyfloatersu: oh wait, so in rust `*=` has lower priority than `-`?  that's nuts :-/

arith/polynomials/src/bivariate/coeff.rs
```diff
@@ -73,21 +72,9 @@ impl<F: PrimeField> BivariatePolynomial<F> {
         ));
 
         // roots of unity for supported_n and supported_m
-        let (omega_0, omega_1) = {
-            let omega = F::ROOT_OF_UNITY;
-            let omega_0 = omega.pow_vartime([(1 << F::S) / self.degree_0 as u64]);
-            let omega_1 = omega.pow_vartime([(1 << F::S) / self.degree_1 as u64]);
+        let omega_0 = primitive_root_of_unity(self.degree_0);
+        let omega_1 = primitive_root_of_unity(self.degree_1);
```
tonyfloatersu, line 77: that's a good one.  I see you stashed some good methods in `utils.rs`

arith/polynomials/src/utils.rs
```diff
@@ -0,0 +1,132 @@
...
+
+#[inline]
+pub fn primitive_root_of_unity<F: FFTField>(group_size: usize) -> F {
+    let log_size = group_size.trailing_zeros() as usize;
+    assert_eq!(1 << log_size, group_size);
+    F::two_adic_generator(log_size)
+}
+
+#[inline]
+pub fn bitreverse(mut n: usize, l: usize) -> usize {
```
tonyfloatersu, line 30: note taking: from prior `arith/polynomials/src/bivariate/bi_fft.rs`

arith/polynomials/src/univariate.rs
```diff
@@ -0,0 +1,35 @@
...
+    }
+}
+
+impl<F: FFTField> From<&UnivariatePolynomial<F>> for UnivariateLagrangePolynomial<F> {
+    #[inline]
+    fn from(poly: &UnivariatePolynomial<F>) -> Self {
+        let coeffs = poly.interpolate();
+        UnivariateLagrangePolynomial::new(coeffs)
+    }
+}
```
tonyfloatersu, line 35: similar structure like `bivariate.rs`

arith/polynomials/src/bivariate/lagrange.rs
```diff
@@ -1,10 +1,34 @@
...
+            numerator * denominator.inv().unwrap()
```
tonyfloatersu: nit: I suppose this can also be `batch_inversed`, once we get all the `denominators` by the end of computation

arith/polynomials/src/univariate/lagrange.rs
```diff
@@ -0,0 +1,37 @@
...
+        for i in 0..n {
+            let mut term = self.coefficients[i];
+            let x_i = omega.exp(i as u128);
+
+            // Compute lagrange basis polynomial
+            // TODO: batch inversion
+            for j in 0..n {
+                if i != j {
+                    let x_j = omega.exp(j as u128);
+                    term *= (*x - x_j) * (x_i - x_j).inv().unwrap();
```
tonyfloatersu, line 31: reminder: this looks like `bivariate/lagrange.rs`'s `lagrange_coefficients`, and batch inverse can be of some use there and here.

arith/polynomials/src/utils.rs
```diff
@@ -0,0 +1,132 @@
...
+//
+// Jolt's ack:
+//
+// Inspired by: https://github.com/TheAlgorithms/Rust/blob/master/src/math/gaussian_elimination.rs
+// Gaussian Elimination of Quadratic Matrices
+// Takes an augmented matrix as input, returns vector of results
+// Wikipedia reference: augmented matrix: https://en.wikipedia.org/wiki/Augmented_matrix
+// Wikipedia reference: algorithm: https://en.wikipedia.org/wiki/Gaussian_elimination
+
+pub fn gaussian_elimination<F: Field>(matrix: &mut [Vec<F>]) -> Vec<F> {
```
tonyfloatersu, line 78: hope I'm not opinionated, but what is the use of introducing gaussian elimination here?

the one use case I have in my mind is using gaussian elimination to interpolate a high degree sumcheck response, and thus we can get the univariate poly's coefficient...  what's your idea about it, on how it fit into the context?

arith/polynomials/src/univariate/coeff.rs
```diff
@@ -0,0 +1,111 @@
...
+            .fold(F::ZERO, |acc, (c, x_i)| acc + *c * *x_i)
+    }
+
+    pub fn from_evals(evals: &[F]) -> Self {
+        Self {
+            coefficients: Self::vandermonde_interpolation(evals),
+        }
+    }
+
+    fn vandermonde_interpolation(evals: &[F]) -> Vec<F> {
```
tonyfloatersu, line 48: oh boy, the runtime on this one would be huge.

arith/polynomials/src/univariate/fft.rs
```diff
@@ -0,0 +1,113 @@
...
+/// The result is that the vector `a`, when
+/// interpreted as the coefficients of a polynomial of degree $n - 1$, is
+/// transformed into the evaluations of this polynomial at each of the $n$
+/// distinct powers of $\omega$. This transformation is invertible by providing
+/// $\omega^{-1}$ in place of $\omega$ and dividing each resulting field element
+/// by $n$.
+///
+/// This will use multithreading if beneficial.
+pub fn best_fft<F: FFTField>(a: &mut [F], omega: F, log_n: u32) {
+    let threads = rayon::current_num_threads();
```
tonyfloatersu, line 20: one of the biggest confusion in this codebase for me is, what is the best (or recommended) practice in using `rayon` or other multithreading libs?

in my context, if I am working on a PCS, then eventually I will be thinking about using multithreading to accelerate the commit time, e.g., in MT.

but take a step back, on a large scope, will multithreading be applied globally, or it will be locally limited in distinct modules, e.g., sumcheck locally use multithreading, kzg-fft locally use multithreading, or hash-based pcs locally use multithreading, yet non of them step on each other, or perf degrading will much likely happen...

arith/polynomials/src/univariate/coeff.rs
tonyfloatersu: now it feels to me that the gaussian elim and vandermonde interpolation is much like to serve the purpose of sumcheck resp interpolation

arith/polynomials/src/univariate/coeff.rs
```diff
@@ -0,0 +1,111 @@
...
+        Self::new(coefficients)
+    }
+
+    #[inline]
+    /// Compute poly / (x-point) using univariate division
+    pub fn quotient_polynomial(&self, point: &F) -> Self {
+        Self::new(univariate_quotient(&self.coefficients, point))
+    }
+
+    /// evaluate the polynomial at (x, y)
```
tonyfloatersu: nit: evaluate at just x? it is a univariate poly

arith/polynomials/src/univariate.rs
```diff
@@ -0,0 +1,35 @@
...
+use arith::FFTField;
+
+#[derive(Clone, Debug, Default, PartialEq, Eq)]
+pub struct UnivariatePolynomial<F> {
+    pub coefficients: Vec<F>,
+}
+
+#[derive(Clone, Debug, Default, PartialEq, Eq)]
+pub struct UnivariateLagrangePolynomial<F> {
+    pub coefficients: Vec<F>,
```
tonyfloatersu, line 19: nit: I have a renaming request, similar to the ones in `bivariate.rs`, that `(Uni/Bi)variateLagrangePolynomial`'s `coefficients` are actually evaluations of the polynomials over the nth-primitive-root-of-unity, and thus `evaluations` makes more sense to me.

I am saying this because the `From` method from `&(Uni/Bi)variatePolynomial` to `(Uni/Bi)variateLagrangePolynomial` invokes a `poly.interpolate`, while both ends of variables are called `coeffs`, which does not make sense to me... that one is coeffs and the other is evals, how come both are coeffs...

arith/polynomials/src/univariate/coeff.rs
```diff
@@ -0,0 +1,111 @@
...
+        let omega = primitive_root_of_unity(self.coefficients.len());
+        let log_n = self.coefficients.len().trailing_zeros();
+        best_fft(&mut res, omega, log_n);
+        res
+    }
+}
+
+/// Compute poly / (x-point) using univariate division
+pub fn univariate_quotient<F: Field>(poly: &[F], point: &F) -> Vec<F> {
+    let timer = start_timer!(|| format!("Univariate quotient of degree {}", poly.len()));
```
tonyfloatersu: this does not feel so right to me, maybe `poly.len() - 1`?

arith/polynomials/src/univariate/coeff.rs
```diff
@@ -0,0 +1,111 @@
...
+        }
+
+        dividend_pos -= 1;
+        difference -= 1;
+    }
+    coefficients.reverse();
+    coefficients.push(F::from(0u32));
+    end_timer!(timer);
+    coefficients
+}
```
tonyfloatersu, line 111: I understand how this works, but this code makes dividing (X - z) a bit way too complex.

can we just rewrite the whole thing by iterating down from top degree down to degree 1, as eventually you will ignore the constant term, which is the remainder here dividing a low degree (X - z)?

arith/polynomials/src/tests/mle.rs
```diff
@@ -0,0 +1,125 @@
...
+    // for example, with num_vars = 4, x is a binary vector of 4, then
+    //  0 0 0 0 -> (1-r0)   * (1-r1)    * (1-r2)    * (1-r3)
+    //  1 0 0 0 -> r0       * (1-r1)    * (1-r2)    * (1-r3)
+    //  0 1 0 0 -> (1-r0)   * r1        * (1-r2)    * (1-r3)
+    //  1 1 0 0 -> r0       * r1        * (1-r2)    * (1-r3)
+    //  ....
+    //  1 1 1 1 -> r0       * r1        * r2        * r3
+    // we will need 2^num_var evaluations
+
+    // First, we build array for {1 - r_i}
```
tonyfloatersu, line 96: one minor question on my side, so we agreed on the multilinear poly's boolean hypercube eval order being small endian, i.e., as illustrated above?

arith/polynomials/src/bivariate.rs
```diff
@@ -0,0 +1,38 @@
...
+mod coeff;
+
+mod lagrange;
+pub use lagrange::*;
+
+use arith::FFTField;
+
+#[derive(Clone, Debug, Default, PartialEq, Eq)]
+pub struct BivariatePolynomial<F> {
+    pub coefficients: Vec<F>,
```
tonyfloatersu, line 13: note taking: the `coefficients` are ordered by `degree_0` x `degree_1`, where `x` has `degree_0` and `y` has `degree_1`

readme.md
```diff
@@ -72,7 +72,7 @@ cargo run --bin=dev-setup --release
 Command template:
 
 ```sh
-RUSTFLAGS="-C target-cpu=native" cargo run --release --bin gkr -- -f [fr|m31ext3] -t [#threads] -s [keccak|poseidon]
+RUSTFLAGS="-C target-cpu=native" cargo run --release --bin gkr -- -f [fr|m31ext3|babybearext3|babybearext4] -t [#threads] -s [keccak|poseidon]
```
tonyfloatersu, line 75: nit: actually, since we already add `-C target-cpu=native` in `.cargo/config.toml`, I guess we no longer need the prefix of `RUSTFLAGS`?

poseidon/src/babybear.rs
```diff
@@ -0,0 +1,142 @@
...
+    pub round_constants: Vec<BabyBearx16>,
+}
+
+impl PoseidonBabyBearParams {
+    #[inline]
+    pub fn new(mut rng: impl RngCore) -> Self {
+        let full_rounds = 8;
+
+        let partial_rounds = 14;
+
```
tonyfloatersu, line 30: I suppose the partial round can be 13 by https://github.com/Plonky3/Plonky3/blob/main/poseidon2/src/round_numbers.rs#L41-L42?  (But that is for Poseidon2, dunno if it applies here)

tree/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "tree"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+arith = { path = "../arith" }
+babybear = { path = "../arith/babybear" }
```
tonyfloatersu, line 8: move babybear to dev-dependencies?  it is only used in test

Cargo.toml
```diff
@@ -31,6 +34,13 @@ halo2curves = { git = "https://github.com/PolyhedraZK/halo2curves", default-feat
 ] }
 itertools = "0.13"
 log = "0.4"
+p3-baby-bear = { git = "https://github.com/Plonky3/Plonky3.git", version = "0.1.0", features = ["nightly-features"] }
+p3-dft = { git = "https://github.com/Plonky3/Plonky3.git", package = "p3-dft" }
+p3-field = { git = "https://github.com/Plonky3/Plonky3.git", version = "0.1.0" }
+p3-matrix = { git = "https://github.com/Plonky3/Plonky3.git", package = "p3-matrix" }
```
tonyfloatersu, line 40: I don't think we need `p3-matrix` here?

Cargo.toml
```diff
@@ -31,6 +34,13 @@ halo2curves = { git = "https://github.com/PolyhedraZK/halo2curves", default-feat
...
+# This import should be unnecessary but p3-baby-bear fails to turn on "nightly-features" for this dependency
+p3-monty-31 = { git = "https://github.com/Plonky3/Plonky3.git", version = "0.1.0", features = ["nightly-features"] }
+plonky2_util = "0.2.0"
```
tonyfloatersu, line 43: same here, where do we need this dependency?

Cargo.toml
```diff
@@ -31,6 +34,13 @@ halo2curves = { git = "https://github.com/PolyhedraZK/halo2curves", default-feat
 ] }
 itertools = "0.13"
 log = "0.4"
+p3-baby-bear = { git = "https://github.com/Plonky3/Plonky3.git", version = "0.1.0", features = ["nightly-features"] }
+p3-dft = { git = "https://github.com/Plonky3/Plonky3.git", package = "p3-dft" }
```
tonyfloatersu, line 38: hmm do we import `p3-dft` for basefold?

tree/src/leaf.rs
```diff
@@ -0,0 +1,66 @@
...
+use sha2::{Digest, Sha512};
+
+use crate::Node;
+
+/// Represents a leaf in the Merkle tree, containing 64 bytes of data stored in a BabyBearx16.
+#[derive(Debug, Copy, Clone, PartialEq, Default)]
+pub struct Leaf<F: Field + FieldSerde> {
+    pub data: F,
+}
+
```
tonyfloatersu, line 14: I have some thoughts here - as you mentioned here, the leaf contains 64 bytes of data, yet the type of leaf is `F: Field + FieldSerde`.

The reason why I wrote the octopos tree hash leaf from a chunk of bytes (https://github.com/cysic-labs/jolt-b/blob/master/octopos/src/hash.rs#L107) is that, I wanted to know explicitly how much data (oblivious of what type it is) can be written in the space, and given that we know what type the bytes should be deserialized to, it would be just fine to deserialize from the leaf.

Yet in this situation, though the field is generic, we somehow assume that the field serialize to 64 bytes, that has to be either Babybear_x16 or M31_x16.

What if we want to hash Orion codeword that works over GF(2) that is binary, then we probably need to construct some 512-length module of GF(2) to fit into the leaf, which seems a bit inconvenient here.

tree/src/tree.rs
```diff
@@ -0,0 +1,233 @@
...
+            .iter()
+            .map(|&leaf| Leaf { data: leaf })
+            .collect::<Vec<Leaf<F>>>();
+        Self::new_with_leaves(leaves)
+    }
+
+    /// Builds a tree with the given leaves.
+    #[inline]
+    pub fn new_with_leaves(leaves: Vec<Leaf<F>>) -> Self {
+        let tree_height = log2(leaves.len() + 1);
```
tonyfloatersu, line 52: i felt like this can be wrong - consider 7 leaves, then you need 8 leaves capacity, while the height from 7 leaves is different from the one from 8 leaves

tonyfloatersu (commented):
half way on tree, gkr and sumcheck to be read, dump the review comments for now

tree/src/tree.rs
tonyfloatersu: maybe they are assuming that the leaves are always being po2, s.t., the height is correct in this case

.github/workflows/ci.yml
zhenfeizhang: seems like plonky3 import will fail without those

tonyfloatersu: ah i see, seems a thing related to self maintained CI

arith/polynomials/src/bivariate/lagrange.rs
zhenfeizhang: sounds good. Implemented

arith/polynomials/src/utils.rs
zhenfeizhang: Well... IIRC this is from Jolt in univariate interpolation. I think we can avoid it.

arith/polynomials/src/univariate/coeff.rs
zhenfeizhang: > Well... IIRC this is from Jolt in univariate interpolation. I think we can avoid it.

I think we can and should optimize this part...

arith/polynomials/src/univariate/fft.rs
zhenfeizhang: I think the parallelization needs to be done at the MPI/circuit level for best performance. Parallelizing FFT usually loses performance as you don't ultilize all the cores

arith/polynomials/src/univariate/coeff.rs
zhenfeizhang: sounds good. this was a legacy code from C

Cargo.toml
zhenfeizhang: those will be needed for basefold

Cargo.toml
zhenfeizhang: those will be needed for basefold

Cargo.toml
zhenfeizhang: those will be needed for basefold

Cargo.toml
tonyfloatersu: ok, i will leave them be

Cargo.toml
tonyfloatersu: same as above, leave them be

Cargo.toml
tonyfloatersu: leave them be at this point

sumcheck/src/prover_helper/arbitrary_gate.rs
```diff
@@ -0,0 +1,187 @@
...
+            // evaluate the claimed degree-ell polynomial at r_i
+            e = poly.evaluate(&r_i);
+        }
+
+        (e, r)
+    }
+}
+
+#[cfg(test)]
+mod test {}
```
tonyfloatersu, line 187: unused code?

sumcheck/src/prover_helper/arbitrary_gate.rs
```diff
@@ -0,0 +1,187 @@
...
+            let uni_poly_eval_points: Vec<F> = (0..combined_degree + 1)
+                .map(|poly_i| accum.iter().map(|mle| mle[poly_i]).sum())
+                .collect();
+
+            let round_uni_poly = UnivariatePolynomial::from_evals(&uni_poly_eval_points);
+
+            // append the prover's message to the transcript
+            // round_uni_poly.
+            // transcript.append_field_element(f);
+            println!("round_uni_poly: {:?}", round_uni_poly.coefficients.len());
```
tonyfloatersu, line 104: I guess you want to say something related to the degree of the univariate poly in the round of response in sumcheck?

sumcheck/src/prover_helper/arbitrary_gate.rs
```diff
@@ -0,0 +1,187 @@
...
+
+        // verify that there is a univariate polynomial for each round
+        assert_eq!(self.uni_polys.len(), num_rounds);
+        for i in 0..self.uni_polys.len() {
+            let poly = &self.uni_polys[i];
+
+            // verify degree bound
+            if poly.degree() != degree_bound {
+                panic!("poly degree incorrect")
+            }
```
tonyfloatersu, line 159: can just `assert_eq!` without using panic?

sumcheck/src/prover_helper/arbitrary_gate.rs
```diff
@@ -0,0 +1,187 @@
...
+    /// - `claim`: Claimed sumcheck evaluation (note: currently unused)
+    /// - `num_rounds`: Number of rounds of sumcheck, or number of variables to bind
+    /// - `polys`: Dense polynomials to combine and sumcheck
+    /// - `comb_func`: Function used to combine each polynomial evaluation
+    /// - `transcript`: Fiat-shamir transcript
+    ///
+    /// Returns (SumcheckInstanceProof, r_eval_point, final_evals)
+    /// - `r_eval_point`: Final random point of evaluation
+    /// - `final_evals`: Each of the polys evaluated at `r_eval_point`
+    pub fn prove_arbitrary<Func, T>(
```
tonyfloatersu, line 29: a minor question regarding the naming - if I am guessing correctly, the naming suggests that it applies to arbitrary gate, i.e., the `comb_func` here.  In a sense, this also suggests specialized sumchecks will be (or have been) implemented for certain cases, say some low degree common cases.   Do these guesses sound reasonable to you?

sumcheck/src/prover_helper/arbitrary_gate.rs
```diff
@@ -0,0 +1,187 @@
...
+                        polys.iter().map(|poly| poly.coeffs[poly_term_i]).collect();
+                    accum[0] += comb_func(&params_zero);
+
+                    // TODO(#28): Can be computed from prev_round_claim - eval_point_0
+                    let params_one: Vec<F> = polys
+                        .iter()
+                        .map(|poly| poly.coeffs[mle_half + poly_term_i])
+                        .collect();
+                    accum[1] += comb_func(&params_one);
+
```
tonyfloatersu, line 69: it seems, the sumcheck summing sequence is from right to left, i.e., sumcheck reduces variables from the rightmost one to the leftmost one.

sumcheck/src/prover_helper/arbitrary_gate.rs
```diff
@@ -0,0 +1,187 @@
...
+                    // D_n(index, r) = D_{n-1}[half + index] + r * (D_{n-1}[half + index] - D_{n-1}[index])
+                    // D_n(index, 0) = D_{n-1}[LOW]
+                    // D_n(index, 1) = D_{n-1}[HIGH]
+                    // D_n(index, 2) = D_{n-1}[HIGH] + (D_{n-1}[HIGH] - D_{n-1}[LOW])
+                    // D_n(index, 3) = D_{n-1}[HIGH] + (D_{n-1}[HIGH] - D_{n-1}[LOW]) + (D_{n-1}[HIGH] - D_{n-1}[LOW])
+                    // ...
+                    let mut existing_term = params_one;
+                    for eval_i in accum.iter_mut().take(combined_degree + 1).skip(2) {
+                        let mut poly_evals = vec![F::zero(); polys.len()];
```
tonyfloatersu, line 78: move the declaration of `poly_evals` buffer outside the loop, eventually it gets overwritten each iteration from evaluation point being 2 to `combine_degree + 1`.

tonyfloatersu: actually, why not just directly `existing_term[poly_i] += (poly.coeffs[mle_half + poly_term_i] - poly.coeffs[poly_term_i]`?

also, you can take advantage of the prior `stage_zero` and `stage_one`, so you don't have to search into poly's coeffs for the prior searched coefficients, which may help perf a bit...

sumcheck/src/prover_helper/arbitrary_gate.rs
```diff
@@ -0,0 +1,187 @@
...
+    {
+        let mut r: Vec<F> = Vec::new();
+        let mut uni_polys: Vec<UnivariatePolynomial<F>> = Vec::new();
+
+        for _round in 0..num_rounds {
+            let mle_half = polys[0].coeffs.len() / 2;
+
+            let accum: Vec<Vec<F>> = (0..mle_half)
+                .map(|poly_term_i| {
+                    let mut accum = vec![F::zero(); combined_degree + 1];
```
tonyfloatersu, line 49: hmmm, please consider another name, not conflicting with outside accum - I guess this variable is keeping track of the `comb_func` evaluations over current point, while the summing variable need to be from 0 to combine-degree, s.t., this allows verifier to sample a random challenge and reduce the claimed summation for next round of sumcheck.

back to what I have proposed, please do consider another naming for this variable.

sumcheck/src/prover_helper/arbitrary_gate.rs
```diff
@@ -0,0 +1,187 @@
...
+                    for eval_i in accum.iter_mut().take(combined_degree + 1).skip(2) {
+                        let mut poly_evals = vec![F::zero(); polys.len()];
+                        for poly_i in 0..polys.len() {
+                            let poly = &polys[poly_i];
+                            poly_evals[poly_i] = existing_term[poly_i]
+                                + poly.coeffs[mle_half + poly_term_i]
+                                - poly.coeffs[poly_term_i];
+                        }
+
+                        *eval_i += comb_func(&poly_evals);
```
tonyfloatersu, line 86: please consider for syntax simplicity, that just use assign here? this is the same case with prior `accum[0]` and `accum[1]` that only visited once and assigned once, so please just assign rather than add to the value...

tonyfloatersu (commented):
wrapping up on `arbitrary_gate.res`, some current code here needs some polishing

bi-kzg/src/lib.rs
tonyfloatersu, line 1: a second thought after finishing code reading - can we take out the pcs trait to somewhere, s.t., pcs implementations afterwards can comply with the trait?  #97 has 2 pcs traits in `bi-kzg` and `basefold`, should be just one pcs trait I guess?

sumcheck/src/tests.rs
```diff
@@ -0,0 +1,110 @@
...
+            })
+            .collect();
+
+        let evals: Vec<Fr> = polys.iter().map(|poly| poly.evaluate(&point)).collect();
+        actual_sum += combination_function(&evals);
+    }
+
+    println!("creat proof");
+    // Create the proof
+    let combined_degree = 2; // Sum of two linear polynomials has degree 2
```
tonyfloatersu, line 64: is that right?

sumcheck/src/tests.rs
```diff
@@ -0,0 +1,110 @@
...
+    println!("generate sum");
+    let mut actual_sum = Fr::zero();
+    for i in 0..(1 << num_vars) {
+        let point: Vec<Fr> = (0..num_vars)
+            .map(|j| {
+                if (i >> j) & 1 == 1 {
+                    Fr::one()
+                } else {
+                    Fr::zero()
+                }
```
tonyfloatersu, line 54: maybe just `Fr::from((i >> j) & 1) as u64)`?

PR #137, opened 2024-10-31
Feature: Orion PCS Prototype

tonyfloatersu:
Orion PCS prototype.


arith/gf2/src/gf2x64.rs
tonyfloatersu, line 1: note: I am totally mimicing `gf2x8` implementation, please review with that file when you read this impl.

arith/gf2_128/src/gf2_ext128x8/neon.rs
```diff
@@ -75,7 +75,7 @@ impl Field for NeonGF2_128x8 {
 
     const SIZE: usize = 16 * 8;
 
-    const FIELD_SIZE: usize = 128 * 8; // in bits
+    const FIELD_SIZE: usize = 128; // in bits
```
tonyfloatersu: I think I caught a bug casually, this should be the GF2_ext128 size, not the size of GF2_ext128x8

PR #140, opened 2024-11-06
pcs trait v1

zhiyong1997:


arith/src/serde.rs
```diff
@@ -56,6 +56,31 @@ field_serde_for_integer!(u64, 8);
 field_serde_for_integer!(usize, 8);
 field_serde_for_integer!(u8, 1);
 
+impl<V: FieldSerde> FieldSerde for Vec<V> {
```
tonyfloatersu, line 59: hmm, i guess this is implemented for `RawCommit`?

pcs/src/traits.rs
```diff
@@ -0,0 +1,63 @@
...
+    type Poly: Clone + Debug;
+    type EvalPoint: Clone + Debug;
+
+    type SRS: Clone + Debug + FieldSerde;
+    type PKey: Clone + Debug + From<Self::SRS> + FieldSerde;
+    type VKey: Clone + Debug + From<Self::SRS> + FieldSerde;
+    type Commitment: Clone + Debug + FieldSerde;
+    type Opening: Clone + Debug + FieldSerde;
+
+    fn gen_srs_for_testing(&self, rng: impl RngCore, params: &Self::Params) -> Self::SRS;
```
tonyfloatersu: propose narrower name like `gen_crs`

tonyfloatersu: oh, I guess you are copying from hyperplonk's pcs trait that this crs should not be used in production

pcs/src/traits.rs
```diff
@@ -0,0 +1,63 @@
...
+
+    fn commit(
+        &self,
+        params: &Self::Params,
+        proving_key: &Self::PKey,
+        poly: &Self::Poly,
+    ) -> Self::Commitment;
```
tonyfloatersu, line 23: I actually have a minor thought here: can we separate `Commitment` into `CommitmentWithData` and `Commitment`?

The thinking process proceeds as follows: supposing you are doing a FRI-based pcs or a linear-code tensor-IOP based pcs, then the commitment itself should be just a MT hash, while commitment with data contains all the codewords related.  In this case, the `commit` method should output a `CommitmentWithData` containing all the MTs committing against related codewords.

Take a step back, say we are using a KZG-alike curve based commitment, then `CommitmentWithData` should be equivalent to `Commitment`, as it falls back to the crs.

(Simply just dumping my thoughts here, but I don't expect you do take further action on this - maybe we should just merge in this one and I will do some modifications on top of it.)

tonyfloatersu (commented):
implementation-wise seems reasonable to me, yet I have some thoughts on the trait design for pcs, open for discussion on this but I don't expect actions on this - I can take over and modify the trait later

tonyfloatersu:
btw if we already have (raw) pcs in a standalone project under expander, shall we just remove the old raw pcs under gkr?

arith/src/serde.rs
zhiyong1997: I think it's good to have FieldSerde implemented for Vec, not only for RawCommit.

pcs/src/traits.rs
zhiyong1997: For the name, Yes. And I think `srs` is actually more frequently used than `crs` in latest literature.

PR #142, opened 2024-11-09
Features: Orion implementation related peripheral building blocks

tonyfloatersu:
This PR contains following things in favor of Orion pcs implementation in #137:
- A GF2x64 SimdField implementation over `u64`.
- A `SimdField` implementation for in `GF2_128` for `GF2` elements.
- Refactoring `pack_size()` method in `SimdField` into trait constant `PACK_SIZE: usize`.
- A bug fix for a faulty `FIELD_SIZE` in `NeonGF2_128x8`.
- Clonable Fiat-Shamir transcript, and challenge index into PCP oracle (credit to #133).
- Extending `ExtensionField` with `Mul<Self::BaseField>` in support of multiplication with base field elems (controversial, up for discussion).

### DIFF
--- a/arith/gf2/src/gf2x64.rs
+++ b/arith/gf2/src/gf2x64.rs
@@ -4,14 +4,13 @@ use arith::{Field, FieldSerde, FieldSerdeResult, SimdField};
 
 use super::GF2;
 
-/// A GF2x8 stores 8 bits of data.
 #[derive(Debug, Clone, Copy, Default, PartialEq)]
-pub struct GF2x8 {
-    pub v: u8,
+pub struct GF2x64 {
+    pub v: u64,
 }
 
-impl FieldSerde for GF2x8 {
-    const SERIALIZED_SIZE: usize = 1;
+impl FieldSerde for GF2x64 {
+    const SERIALIZED_SIZE: usize = 8;
 
     #[inline(always)]
     fn serialize_into<W: std::io::Write>(&self, mut writer: W) -> FieldSerdeResult<()> {
@@ -23,41 +22,38 @@ impl FieldSerde for GF2x8 {
     fn deserialize_from<R: std::io::Read>(mut reader: R) -> FieldSerdeResult<Self> {
         let mut u = [0u8; Self::SERIALIZED_SIZE];
         reader.read_exact(&mut u)?;
-        Ok(GF2x8 { v: u[0] })
+        Ok(GF2x64 {
+            v: u64::from_le_bytes(u),
+        })
     }
 
     #[inline]
-    fn try_deserialize_from_ecc_format<R: std::io::Read>(mut _reader: R) -> FieldSerdeResult<Self> {
-        unimplemented!("We don't have serialization in ecc for gf2x8")
-        // let mut u = [0u8; 32];
-        // reader.read_exact(&mut u)?;
-        // Ok(GF2x8 { v: u[0] })
+    fn try_deserialize_from_ecc_format<R: std::io::Read>(_reader: R) -> FieldSerdeResult<Self> {
+        unimplemented!("We don't have serialization in ecc for gf2x64")
     }
 }
 
-impl Field for GF2x8 {
-    // still will pack 8 bits into a u8
+impl Field for GF2x64 {
+    const NAME: &'static str = "Galois Field 2 SIMD 64";
 
-    const NAME: &'static str = "Galios Field 2 SIMD 8";
+    const SIZE: usize = 8;
 
-    const SIZE: usize = 1;
+    const FIELD_SIZE: usize = 1;
 
-    const FIELD_SIZE: usize = 1; // in bits
+    const ZERO: Self = GF2x64 { v: 0 };
 
-    const ZERO: Self = GF2x8 { v: 0 };
+    const ONE: Self = GF2x64 { v: !0u64 };
 
-    const ONE: Self = GF2x8 { v: 255 };
-
-    const INV_2: Self = GF2x8 { v: 0 }; // should not be used
+    const INV_2: Self = GF2x64 { v: 0 }; // NOTE: should not be used
 
     #[inline(always)]
     fn zero() -> Self {
-        GF2x8 { v: 0 }
+        GF2x64::ZERO
     }
 
     #[inline(always)]
     fn one() -> Self {
-        GF2x8 { v: 255 }
+        GF2x64::ONE
     }
 
     #[inline(always)]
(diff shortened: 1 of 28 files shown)


arith/gf2_128/src/gf2_ext128/avx.rs
```diff
@@ -318,3 +328,41 @@ fn mul_internal(a: &AVXGF2_128, b: &AVXGF2_128) -> AVXGF2_128 {
         v: unsafe { gfmul(a.v, b.v) },
     }
 }
+
+impl SimdField for AVXGF2_128 {
```
zhiyong1997: There might be some misunderstanding here. AVXGF2_128 refers to the degree-128 extension of GF2. It is not an SIMD field.

zhiyong1997: We should probably rename it to something like AVXGF2Ext128.

arith/gf2_128/src/gf2_ext128/neon.rs
```diff
@@ -394,3 +403,41 @@ pub(crate) fn mul_by_x_internal(a: &uint32x4_t) -> uint32x4_t {
         vreinterpretq_u32_u64(res)
     }
 }
+
+impl SimdField for NeonGF2_128 {
```
zhiyong1997: Same problem here. SimdField should not be implemented for this.

arith/gf2_128/src/gf2_ext128/avx.rs
tonyfloatersu: fair point, I was being lazy on this by directly taking advantage of `AVXGF2_128`, which can be viewed as a wrapper around `__m128i` - maybe we should just separate out a standalone thing like `AVXGF2x128` rather than the status quo...

arith/gf2/src/gf2x64.rs
```diff
@@ -0,0 +1,301 @@
...
+        GF2x64 { v: rng.next_u64() }
+    }
+
+    #[inline(always)]
+    fn random_bool(mut rng: impl rand::RngCore) -> Self {
+        GF2x64 { v: rng.next_u64() }
+    }
+
+    #[inline(always)]
+    fn exp(&self, exponent: u128) -> Self {
```
zhiyong1997, line 75: I think `exp` should simply return `*self`, regardless of what `exponent` we have. The same problem exists for previous implementation of `exp` in `GF2x8`.

arith/gf2_128/src/tests.rs
```diff
@@ -13,7 +13,8 @@ use crate::{GF2_128x8, GF2_128};
 
 #[test]
 fn test_simd_field() {
-    random_simd_field_tests::<GF2_128x8>("Simd GF2 Ext128".to_string());
+    random_simd_field_tests::<GF2_128>("Simd for GF2 over GF2Ext128".to_string());
```
zhiyong1997: The same SIMD field problem here.

arith/src/extension_field.rs
```diff
@@ -1,10 +1,14 @@
...
+
 use crate::{Field, FieldSerde};
 
 /// Configurations for Extension Field over
 /// - either the Binomial polynomial x^DEGREE - W
 /// - or the AES polynomial x^128 + x^7 + x^2 + x + 1
 //
-pub trait ExtensionField: From<Self::BaseField> + Field + FieldSerde {
+pub trait ExtensionField:
+    Mul<Self::BaseField> + From<Self::BaseField> + Field + FieldSerde
```
zhiyong1997, line 10: I think it's better to have `Add` as well if we have `Mul` here, but I'm fine with this.

zhiyong1997 (changes requested):
As commented.

arith/gf2/src/gf2x64.rs
tonyfloatersu: why? the exp should align with gf2 impl, as it reflects simd - I don't get why each cell of gf2 maps to self regardless of exponent. 

elaborate more.

zhiyong1997: In gf2, we have $0\times 0 = 0, 1\times 1 = 1$. As a result, for any $x \in \{0, 1\}, x^e = x$. I believe we implemented this incorrectly for `gf2` and `gf2x8`.

tonyfloatersu: so long as e is non zero.

zhiyong1997: oh, you're right...

tonyfloatersu: ok - we are both right partly! the implementation is still wrong in a way - that the condition should branch into 2 conditioned that exponent is 0 or not - I will work on it, now I am focusing on neon thing...

zhiyong1997: The condition is a little confusing. It seems $0^2$ would equal $1$ without any restriction of the exponent.

zhiyong1997: How about changing `exponent%2 == 0` to `exponent==0`

zhiyong1997: Or probably it's ok to simply return `*self` since $0^0$ is not well-defined anyway. I don't know.

arith/gf2_128/src/tests.rs
tonyfloatersu: reverted in https://github.com/PolyhedraZK/Expander/pull/142/commits/4fd26abc44b17fa828b34d4bccf3a2eba86029d4

arith/gf2_128/src/gf2_ext128/avx.rs
tonyfloatersu: reverted by https://github.com/PolyhedraZK/Expander/pull/142/commits/4fd26abc44b17fa828b34d4bccf3a2eba86029d4

arith/gf2_128/src/gf2_ext128/neon.rs
tonyfloatersu: reverted by https://github.com/PolyhedraZK/Expander/pull/142/commits/4fd26abc44b17fa828b34d4bccf3a2eba86029d4

arith/gf2/src/gf2x64.rs
tonyfloatersu: see https://github.com/PolyhedraZK/Expander/pull/142/commits/33e9d8cc26ea048eb1be0e88ebd67d1345615ad4 - I am trying to cover all possible cases here...

zhiyong1997 (approved):

PR #143, opened 2024-11-09
Feature: Merkle Tree implementation as Orion pcs building block

tonyfloatersu:
Credits to #133, with merkle tree leaves being compact field elements

### DIFF
--- a/tree/src/tree.rs
+++ b/tree/src/tree.rs
@@ -86,17 +86,13 @@ impl Tree {
         F: Field + FieldSerde,
         PackF: SimdField<Scalar = F>,
     {
-        let serialized_bytes: Vec<_> = field_elems
-            .iter()
-            .map(|&p| {
-                let mut buffer = Vec::new();
-                p.serialize_into(&mut buffer).unwrap();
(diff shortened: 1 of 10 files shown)


tree/src/tree.rs
```diff
@@ -0,0 +1,356 @@
...
+    /// Create a tree with compact serialization of field elements as leaves,
+    /// by taking advantage of PackF
+    #[inline]
+    pub fn compact_new_with_field_elems<F, PackF>(field_elems: &[F]) -> Self
+    where
+        F: Field + FieldSerde,
+        PackF: SimdField<Scalar = F>,
+    {
+        let packed_elems: Vec<PackF> = field_elems
+            .chunks(PackF::pack_size())
```
tonyfloatersu: should change back to `PackF::PACK_SIZE` after #142

zhiyong1997 (commented):
LGTM. It might be helpful to have batch proof for non-consecutive leaves, which we can implement in the future.

zhiyong1997 (approved):

PR #145, opened 2024-11-13
Minor: SIMD GF2 packing

tonyfloatersu:


tonyfloatersu:
hmm the perf is not as good as I anticipated... I keep it myself and maybe serve this one later...

PR #150, opened 2024-11-19
PCS Trait V2 & Integration

zhiyong1997:


polynomial_commitment_scheme/src/traits.rs
tonyfloatersu, line 1: The trait definition looks generally good to me, I have one (I promise the last one) proposition - consider IOP based PCSs that relies on Fiat-Shamir to be non-interactive, so can have have space of passing transcript in both `open` and `verify`? That should work for orion side.

polynomial_commitment_scheme/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "polynomial_commitment_scheme"
```
tonyfloatersu: my apologies for suggesting a slightly shorter naming (again): consider "poly_commit"? just suggesting a medium width naming here...

gkr/src/exec.rs
```diff
@@ -103,9 +119,23 @@ async fn run_command<'a, Cfg: GKRConfig>(
...
             let (proof, claimed_v) =
                 load_proof_and_claimed_v(&bytes).expect("Unable to deserialize proof.");
+
+            // TODO: Read PCS  setup from files
+            let (pcs_params, _pcs_proving_key, pcs_verification_key, mut _pcs_scratch) =
+                expander_pcs_init_unsafe::<Cfg::FieldConfig, Cfg::PCS>(
+                    circuit.log_input_size(),
+                    &config.mpi_config,
+                );
             let verifier = gkr::Verifier::new(&config);
```
tonyfloatersu: I think you managed to verify here is sheer luck - as you are only relying on raw commitment with no randomness going around.

I guess we don't have enough facilities to make it work correctly at this point, but I need to keep track of this and clean this up later.

polynomial_commitment_scheme/src/lib.rs
```diff
@@ -0,0 +1,44 @@
...
+};
+
+#[allow(clippy::type_complexity)]
+pub fn expander_pcs_init_unsafe<
+    FieldConfig: GKRFieldConfig,
+    PCS: PCSForExpanderGKR<FieldConfig>,
+>(
+    n_input_vars: usize,
+    mpi_config: &MPIConfig,
+) -> (
```
tonyfloatersu: I don't think this is quite an API for production, as the randomness consumption is ad hoc in the method... I am also keeping track of this and will traverse back later on it.

tonyfloatersu (commented):
I am done with the first pass, and leaning towards merging with minimal changes, as I see this being helpful, yet somehow more work need to be done to integrate Orion into a full e2e, e.g., reading prover/verifier-key from a serialized file format, and some work in config macro.

polynomial_commitment_scheme/Cargo.toml
zhiyong1997: Will be editted in next commit.

polynomial_commitment_scheme/src/traits.rs
zhiyong1997: This makes sense to me.

gkr/src/exec.rs
zhiyong1997: That's true. I'm aware of this as well.

gkr/src/prover/linear_gkr.rs
```diff
@@ -99,28 +109,76 @@ impl<Cfg: GKRConfig> Prover<Cfg> {
...
+    ) {
+        let opening = Cfg::PCS::open(
+            pcs_params,
+            &self.config.mpi_config,
+            pcs_proving_key,
+            input_layer_mle,
+            open_at,
+            transcript,
+            pcs_scratch,
+        );
```
tonyfloatersu, line 179: nit: this is the only call site of `PCS::open`, if you feel concerned about the transcript appending and Fiat-Shamir appending, maybe it is the place you write transcript lock proof and release lock.

tonyfloatersu (commented):
almost there, only one comment

tonyfloatersu (approved):
approving for now, I take care of it afterwards

PR #153, opened 2024-11-21
Feature: Orion PCS Integration

tonyfloatersu:
- [x] Orion (base field / simd field) impls
- [x] Orion MPI functionality (proof aggregating from different parties)
- [x] MPI environment spin up for a real e2e testing
- [x] integrating PCS into GKR configs


arith/src/simd_field.rs
```diff
@@ -14,6 +14,18 @@ pub trait SimdField: From<Self::Scalar> + Field {
...
     fn pack(base_vec: &[Self::Scalar]) -> Self;
 
+    /// pack a vec of simd field with same scalar field into self
+    fn pack_from_simd<PackF>(simd_vec: &[PackF]) -> Self
+    where
+        PackF: SimdField<Scalar = Self::Scalar>,
+    {
+        assert_eq!(simd_vec.len() * PackF::PACK_SIZE, Self::PACK_SIZE);
+        let mut temp: Vec<_> = simd_vec.to_vec();
+        temp.reverse();
```
zhiyong1997: Is this for `gf2x128` and `gf2x64` only? `reverse` here is a little confusing.

poly_commit/src/traits.rs
```diff
@@ -143,3 +168,37 @@ pub trait PCSForExpanderGKR<C: GKRFieldConfig, T: Transcript<C::ChallengeField>>
         opening: &Self::Opening,
     ) -> bool;
 }
+
+pub(crate) trait TensorCodeIOPPCS {
+    const LEAVES_IN_RANGE_OPENING: usize = 2;
+
+    fn codeword_len(&self) -> usize;
+
+    fn hamming_weight(&self) -> f64;
```
zhiyong1997: From the context, I think this refers to the `relative distance` or `minimal relative hamming weight/distance`.

zhiyong1997 (approved):

zhiyong1997:
I've went throught the base field impl, but haven't touched the `simd-agg`-related code. The protocol itself and the security parameters: `proximity_repetition` and `query_complexity` both look good to me. I'll leave an approve here first, and keep reading the `simd-agg` part to see whether there are any problems introduced by optimization.

poly_commit/src/traits.rs
tonyfloatersu: sure - I will go with `minimum_hamming_weight` suggesting the min non-zero number of alphabets in a codeword in https://github.com/PolyhedraZK/Expander/pull/153/commits/a2f62ae1ee3f4e6d168edb18edd0db6961fbe9a4

arith/src/simd_field.rs
tonyfloatersu: right - this was introduced to pack `gf2x8`s into `gf2x64` or `gf2x128`... the reverse here is to fit into the bit ordering, that allows for unpacking in correct order, but truly, leaving without a note seems a bit confusing

PR #160, opened 2024-12-17
Minor: Poseidon M31 Transcript

tonyfloatersu:


arith/field_hashers/src/poseidon/impls.rs
```diff
@@ -0,0 +1,210 @@
...
+    /// FULL_ROUNDS in a poseidon permutation.
+    const FULL_ROUNDS: usize;
+
+    /// PARTIAL_ROUNDS in a poseidon permutation.
+    const PARTIAL_ROUNDS: usize;
+
+    /// STATE_WIDTH stands for the number of field elements that build up a poseidon state.
+    const STATE_WIDTH: usize;
+
+    /// CAPACITY is the nubmer of output field elements (field here stands for InputF).
```
zhenfeizhang: nubmer -> number

arith/field_hashers/src/poseidon/impls.rs
```diff
@@ -0,0 +1,210 @@
...
+    + Mul<Self, Output = Self>
+    + for<'a> Mul<&'a Self, Output = Self>
+    + MulAssign<Self>
+    + for<'a> MulAssign<&'a Self>
+{
+    /// ElemT is the element (base field) type constructing the poseidon state instance.
+    type ElemT: FieldForECC;
+
+    /// SBOX_POW is the lowest exponential pow for poseidon sbox.
+    const SBOX_POW: usize = compile_time_sbox_alpha::<Self::ElemT>();
```
zhenfeizhang: shall we hard code this in each field's implementation?

arith/field_hashers/src/poseidon/impls.rs
```diff
@@ -0,0 +1,210 @@
...
+                hadamard.to_elems().iter().sum()
+            })
+            .collect::<Vec<_>>();
+
+        *self = Self::from_elems(&res)
+    }
+
+    /// full_round_sbox applies x -> x^(sbox_exp) to all elements in the state.
+    fn full_round_sbox(&mut self) {
+        let mut elts = self.to_elems();
```
zhenfeizhang, line 72: this is fine for trait implementations. but for instantiations, we prob want to overload it so that we don't need to perform two unnecessary clones

transcript/src/transcript.rs
```diff
@@ -175,97 +173,133 @@ impl<F: Field, H: FiatShamirBytesHash> BytesHashTranscript<F, H> {
...
+        let elems = self.generate_challenge_field_elements(base_field_elems_num);
+
+        let mut res = Vec::new();
+        let mut buffer = Vec::new();
+
+        elems.iter().for_each(|t| {
+            t.serialize_into(&mut buffer).unwrap();
+            res.extend_from_slice(&buffer);
+        });
+        res.resize(n_bytes, 0);
```
zhenfeizhang: nits: `res.truncate(n_bytes);`

(should have spotted it earlier)

zhenfeizhang (approved):

arith/field_hashers/src/poseidon/impls.rs
tonyfloatersu: good point - I am just leaving a baseline trait impl for functionality, but truly I should leave an override in M31x16 impl side, see latest commit.

arith/field_hashers/src/poseidon/impls.rs
tonyfloatersu: good catch.

arith/field_hashers/src/poseidon/impls.rs
tonyfloatersu: fair point - I removed the compile time alpha derivation, and leave the trait const being empty and up to implementations and instantiations in the latest commit.

PR #162, opened 2024-12-18
Minor: PCS CRS from external rng

tonyfloatersu:
As discussed in #150 (https://github.com/PolyhedraZK/Expander/pull/150#discussion_r1850938446), we want to allow for  randomness being passed in for PCS init, rather than calling randomness form thread rng in an ad hoc style.  This PR refactored into `gkr` crate and `pcs` crate, resolving this problem. 

This is part of #153 , that is taken out to make review easier.

### DIFF
--- a/gkr/Cargo.toml
+++ b/gkr/Cargo.toml
@@ -28,6 +28,7 @@ sha2.workspace = true
 halo2curves.workspace = true
 thiserror.workspace = true
 ethnum.workspace = true
+rand_chacha.workspace = true
 
 # for the server
 bytes.workspace = true
--- a/gkr/src/prover/linear_gkr.rs
+++ b/gkr/src/prover/linear_gkr.rs
@@ -168,6 +168,7 @@ impl<Cfg: GKRConfig> Prover<Cfg> {
         pcs_scratch: &mut <Cfg::PCS as PCSForExpanderGKR<Cfg::FieldConfig, Cfg::Transcript>>::ScratchPad,
         transcript: &mut Cfg::Transcript,
     ) {
+        transcript.lock_proof();
         let opening = Cfg::PCS::open(
             pcs_params,
             &self.config.mpi_config,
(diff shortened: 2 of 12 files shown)


gkr/Cargo.toml
```diff
@@ -28,6 +28,7 @@ sha2.workspace = true
 halo2curves.workspace = true
 thiserror.workspace = true
 ethnum.workspace = true
+rand_chacha.workspace = true
```
zhenfeizhang, line 31: nits: this is a dev-dependency?

gkr/src/prover/linear_gkr.rs
```diff
@@ -168,6 +168,7 @@ impl<Cfg: GKRConfig> Prover<Cfg> {
         pcs_scratch: &mut <Cfg::PCS as PCSForExpanderGKR<Cfg::FieldConfig, Cfg::Transcript>>::ScratchPad,
         transcript: &mut Cfg::Transcript,
     ) {
+        transcript.lock_proof();
```
zhenfeizhang, line 171: why do you need to lock proof (seems irrelevant to this PR)

zhenfeizhang (approved):
LGTM. just a couple of minor comments

gkr/Cargo.toml
tonyfloatersu: unfortunately - this is not quite a dev-dependency - just to ensure that PCS (in benchmarking binary) are generated with same CRS, this one gets into the dependency

zhenfeizhang: oh i see. sounds good.

gkr/src/prover/linear_gkr.rs
tonyfloatersu: I don't recall the exact answer, but I do remember the context - supposing the PCS is instantiated from some multi-round IOP, WLOG we say FRI as an example to carry on - then the PCS-IOP challenge should be irrelevant of the whole GKR proof, but rather inferred from the Fiat-Shamir transcript.  In this way, the PCS internal IOP challenges (e.g., FRI challenges) should not get into the transcript proof (in this context, the PCS opening), and thus we should add a context-alike thing to lock and unlock the transcript proof, just in case some day we put FRI into the codebase (though we are unsure about that).

PR #165, opened 2024-12-24
Expose expander-exec APIs for ECC

mcfx:
ECC has a built-in Expander to handle simple proof and verification. In recent Expander updates, `expander_config` was split, and structs like `GF2ExtConfigSha2` have become user-defined. To avoid introducing too many dependencies on the expander in ECC and to minimize duplicate code, this PR exposes the proof and verification process in `expander-exec` as a simple API, intended to be called by ECC. The corresponding PR on the ECC side is [ExpanderCompilerCollection#57](https://github.com/PolyhedraZK/ExpanderCompilerCollection/pull/57).


tonyfloatersu (approved):
leaving an approval here - the change on `pub` for GKR config seems fine with me

PR #168, opened 2025-01-02
Minor: Mersenne31 GKR recursive verifier

tonyfloatersu:


zhiyong1997 (approved):
I went over the changes very quickly. The introduce of new fields and transcript looks good to me.

PR #169, opened 2025-01-03
mpi with rayon

zhenfeizhang:


config/mpi_config/src/mpi_config.rs
```diff
@@ -0,0 +1,105 @@
...
+
+    #[inline]
+    pub fn world_size(&self) -> i32 {
+        self.world_size
+    }
+
+    #[inline]
+    /// check the caller's thread is the root thread
+    pub fn is_root(&self) -> bool {
+        rayon::current_thread_index().unwrap() == 0
```
zhiyong1997, line 72: I think `current_thread_index` may not be the actual root. For example, if we start 100 threads with `into_par_iter` to call `current_thread_index` only, 50 of them return 0 and the other 50 return 1 on my laptop.

config/mpi_config/src/mpi_config.rs
```diff
@@ -0,0 +1,105 @@
...
+        let mut pending = (0..total).collect::<Vec<_>>();
+        let mut results: Vec<&[u8]> = vec![&[]; total];
+        let mut wait_cycles = 0;
+
+        // Keep going until we've read from all threads
+        while !pending.is_empty() {
+            // Use retain to avoid re-checking already synced threads
+            pending.retain(|&i| {
+                let len = self.threads[i].size();
+                if len >= end {
```
zhiyong1997, line 142: I think it is in general correct to use the `len` value to check whether new data has been successfully appended to the local buffer. However, I have a concern about the syncronization. In the `append` function of `atomic_vec`, the `len` value seems to have been updated  before the actually data is written, which means some other thread may notice the new value of `len`, and read the dirty data. Will this be a problem?

PR #174, opened 2025-01-13
fix mpi ci

zhenfeizhang:


scripts/install.py
```diff
@@ -3,6 +3,7 @@
 
 if __name__ == "__main__":
     if platform == "darwin": # mac os
-        subprocess.run(["brew", "install", "openmpi"])
+        subprocess.run(["brew", "install", "gcc", "make", "openmpi"])
     else:
-        pass # Do nothing, assuming mpi has already been installed
+        subprocess.run(["sudo", "apt-get", "update"])
+        subprocess.run(["sudo", "apt-get", "install", "-y", "build-essential"])
```
tonyfloatersu: I think this is somewhat beyond the CI workflow fixing - see https://github.com/PolyhedraZK/Expander/actions/runs/12758489448/job/35560625367 that we are attempting to install dependency in one of the CI machines, and it seems to be hanging forever...

PR #190, opened 2025-01-30
fix: fix redundant conversion in assertion in tests.rs

famouswizard:
Noticed a redundant `.into()` in the assertion. Since `EXAMPLE_OUT` is already `[u8; 32]`, there's no need for conversion—it could lead to compilation issues or an unintended comparison.

Replaced it with a direct assertion, making the code cleaner and more correct.


famouswizard:
This way, you compare two byte arrays directly, without unnecessary conversions.

PR #191, opened 2025-01-31
PCS: MultilinearKZG implementation and integration

tonyfloatersu:


poly_commit/src/kzg/kzg_structs.rs
```diff
@@ -0,0 +1,131 @@
...
+impl<E: Engine> From<&CoefFormUniKZGSRS<E>> for UniKZGVerifierParams<E> {
+    fn from(value: &CoefFormUniKZGSRS<E>) -> Self {
+        Self {
+            tau_g2: value.tau_g2,
+        }
+    }
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct CoefFormBiKZGLocalSRS<E: Engine> {
```
zhenfeizhang: what does `Local` mean in this context?

poly_commit/src/kzg/kzg_structs.rs
```diff
@@ -0,0 +1,131 @@
...
+    pub tau_x_srs: CoefFormUniKZGSRS<E>,
+    pub tau_y_srs: CoefFormUniKZGSRS<E>,
+}
+
+/// Bivariate KZG PCS verifier's params.
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Default)]
+pub struct BiKZGVerifierParam<E: Engine> {
+    /// tau_0 over G2.
```
zhenfeizhang: `tau_x`?

poly_commit/src/kzg/pcs_for_expander.rs
```diff
@@ -0,0 +1,137 @@
...
+
+        let mut root_gathering_commits: Vec<E::G1Affine> =
+            vec![E::G1Affine::default(); mpi_config.world_size()];
+        mpi_config.gather_vec(&vec![local_commitment], &mut root_gathering_commits);
+
+        let mut final_commit = E::G1Affine::default();
+        if mpi_config.is_root() {
+            final_commit = root_gathering_commits
+                .iter()
+                .map(|c| c.to_curve())
```
zhenfeizhang: minor: `coeff_form_uni_kzg_commit` may return a `Projective` and here you can use `batch_normalization`

poly_commit/src/kzg/pcs_for_expander.rs
```diff
@@ -0,0 +1,137 @@
...
+        {
+            let mut serialized_commitment: Vec<u8> = Vec::new();
+            final_commit
+                .serialize_into(&mut serialized_commitment)
+                .unwrap();
+
+            mpi_config.root_broadcast_bytes(&mut serialized_commitment);
+            final_commit = {
+                let mut cursor = Cursor::new(serialized_commitment);
+                E::G1Affine::deserialize_from(&mut cursor).unwrap()
```
zhenfeizhang: here we are paying the cost of (de)serialization.... there is an `on_curve()` check within deserialization which is quite costly. 
~I think we can send `(x, y)` directly, something like~

``` rust
            let cord = final_commit.coordinates().unwrap();
            let x = cord.x();
            let y = cord.y();
            // omitted: serialize the coordinates and sync them via MPI
            let p = <E::G1Affine as CurveAffine>::from_xy(*x, *y).unwrap();
```

edit: this is actually worse... `from_xy()` also does an `on_curve()` check. I guess there isn't much we can do here unless we use unsafe casting. so let's keep with your method.

poly_commit/src/kzg/kzg_structs.rs
```diff
@@ -0,0 +1,131 @@
...
+impl<E: Engine> StructuredReferenceString for CoefFormBiKZGLocalSRS<E>
+where
+    <E as Engine>::G1Affine: FieldSerde,
+    <E as Engine>::G2Affine: FieldSerde,
+{
+    type PKey = CoefFormBiKZGLocalSRS<E>;
+    type VKey = BiKZGVerifierParam<E>;
+
+    fn into_keys(self) -> (Self::PKey, Self::VKey) {
+        (self.clone(), From::from(&self))
```
zhenfeizhang: it is quite expensive to clone the SRS -- especially if every thread needs to clone their own SRS.

 we prob want to use `Arc` and the `PKey` would be a pointer to the SRS?

poly_commit/src/kzg/utils.rs
```diff
@@ -0,0 +1,170 @@
...
+        powers.push(cur);
+        cur *= x;
+    }
+    powers
+}
+
+/// Given a univariate polynomial of coefficient form f(X) = c0 + c1 X + ... + cn X^n
+/// and perform the division f(X) / (X - \alpha).
+#[inline(always)]
+pub(crate) fn univariate_degree_one_quotient<F: Field>(coeffs: &[F], alpha: F) -> (Vec<F>, F) {
```
zhenfeizhang, line 20: maybe move some of those helper functions to `arith/polynomials` and create a `univariate.rs`?

tonyfloatersu: right, I introduced #225 partly for this, maybe I should move more of these computation to the univariate poly implementation

zhenfeizhang: sounds good

poly_commit/src/kzg/kzg_structs.rs
```diff
@@ -0,0 +1,131 @@
...
+        Self {
+            powers_of_tau: Vec::default(),
+            tau_g2: E::G2Affine::default(),
+        }
+    }
+}
+
+impl<E: Engine> StructuredReferenceString for CoefFormUniKZGSRS<E>
+where
+    <E as Engine>::G1Affine: FieldSerde,
```
zhenfeizhang: will need to change to `ExpSerde` as #214 is merged

tonyfloatersu: good point, see https://github.com/PolyhedraZK/Expander/pull/191/commits/41e4c9f9927f3337fcd9a85cf6c13ec5309a0783

poly_commit/src/kzg/kzg_structs.rs
zhenfeizhang: nvm... i see that `local` refers to the `x` variable for which you compute locally

poly_commit/src/kzg/kzg_structs.rs
tonyfloatersu: right, should be `tau_x` - this must be me didn't align the naming from earlier development

poly_commit/src/kzg/kzg_structs.rs
tonyfloatersu: good point, I suppose we should use a pointer thing to avoid the whole deep copy

poly_commit/src/kzg/hyper_bikzg.rs
```diff
@@ -0,0 +1,649 @@
...
+        let mut folded_x_coms_g1 = vec![g1_zero; local_folded_x_oracle_commits.len()];
+
+        root_gathering_folded_oracle_commits
+            .chunks(local_folded_x_oracle_commits.len())
+            .for_each(|folded_oracles| {
+                izip!(&mut folded_x_coms_g1, folded_oracles)
+                    .for_each(|(x_com_i, oracle_i)| *x_com_i += oracle_i.to_curve())
+            });
+
+        folded_x_oracle_commits = folded_x_coms_g1.iter().map(|f| f.to_affine()).collect();
```
zhenfeizhang: batch normalization

poly_commit/src/kzg/hyper_kzg_structs.rs
```diff
@@ -0,0 +1,362 @@
...
+use crate::*;
+
+#[derive(Clone, Debug)]
+pub struct HyperKZGExportedLocalEvals<E: Engine> {
+    pub beta_x2_eval: E::Fr,
+    pub pos_beta_x_evals: Vec<E::Fr>,
+    pub neg_beta_x_evals: Vec<E::Fr>,
+}
+
+impl<E: Engine> Default for HyperKZGExportedLocalEvals<E> {
```
zhenfeizhang: hmmm.. I see there are a few impl here because `Engine` does not impl `Default`.

one alternative way i used to use is

```
#[derive(Clone, Debug, Derivative)]
#[derivative(Default(bound=""))]
pub(crate) struct HyperKZGLocalEvals<E: Engine> {
    pub(crate) beta2_evals: Vec<E::Fr>,
    pub(crate) pos_beta_evals: Vec<E::Fr>,
    pub(crate) neg_beta_evals: Vec<E::Fr>,
}
``` 


this requires `derivative = "2.2"` package... and it allows for bypassing the trait bound check for E on default...

// just FYI, and I don't have a preference one way or another

tonyfloatersu: sure, I will take on this one, I recall that the Engine Default thing is a quite pain in the butt

zhenfeizhang (approved):

poly_commit/src/kzg/kzg_structs.rs
```diff
@@ -34,7 +34,8 @@ where
     type VKey = UniKZGVerifierParams<E>;
 
     fn into_keys(self) -> (Self::PKey, Self::VKey) {
-        (self.clone(), From::from(&self))
+        let vk: Self::VKey = From::from(&self);
```
zhenfeizhang, line 37: ha... this is easier than I thought

PR #195, opened 2025-02-06
reduce input claims from 2 to 1 for non-Raw PCS

zhiyong1997:


circuit/src/layered/circuit.rs
```diff
@@ -3,16 +3,17 @@ use std::io::Cursor;
...
 use crate::*;
 
 #[derive(Debug, Clone, Default)]
 pub struct StructureInfo {
-    // this var name is a bit misleading -- the power5 gate's max degree is > 1.
-    // this is really try to see if we need to use two phase GKR or not
-    pub max_degree_one: bool,
+    // If a layer contains only linear combination of fan-in-one gates, we can skip the second
+    // phase of sumcheck e.g. y = a + b + c, and y = a^5 + b^5 + c^5
+    pub skip_sumcheck_phase_two: bool,
```
tonyfloatersu, line 16: minor: Are you referring to the fact that - on the input layer, conditioned that there exists mul gates, then there would be a phase 2 opening, i.e., a 2nd PCS opening, and that reflects on naming that the phase 2 sumcheck opening is skipped?

circuit/src/layered/circuit.rs
```diff
@@ -293,4 +310,31 @@ impl<C: GKRFieldConfig> Circuit<C> {
...
+                coef: C::CircuitField::ONE,
+                coef_type: CoefType::Constant,
+                gate_type: 0,
+            });
+        }
+
+        input_relay_layer.structure_info.skip_sumcheck_phase_two = true;
+
+        self.layers.insert(0, input_relay_layer);
+    }
```
tonyfloatersu, line 339: I see, I somehow realized that, supposing that mul gate exists against input layer, then there does exists 2 PCS opening.

We now introduce a layer with wiring predicate being simply mapping the input to the layer above, then in GKR internal layers (not input layer), the 2 opening claims is randomly merged together with a random linear combination, and thus the "relayed" input layer contains only addition gate with fan in 1, and thus only 1 PCS opening.  I guess I am good with this, am I approaching it in the right direction?

circuit/src/layered/circuit.rs
```diff
@@ -142,9 +143,25 @@ impl<C: GKRFieldConfig> Clone for Circuit<C> {
...
+        // If there will be two claims for the input
+        // Introduce an extra relay layer before the input layer
+        if !circuit.layers[0].structure_info.skip_sumcheck_phase_two {
+            match Cfg::PCS_TYPE {
+                // Raw PCS costs nothing in opening, so no need to add relay layer
+                // But we can probably add it in the future for verifier's convenience
+                config::PolynomialCommitmentType::Raw => (),
+                _ => circuit.add_input_relay_layer(),
+            }
+        }
```
tonyfloatersu, line 162: I have a random thought, that can we just clean up the information of skipping phase 2, and on default only do 1 PCS opening?  An additional layer of sumcheck would be generally cheaper than a PCS opening, but I understand that removing the condition of skipping phase 2 means a few changes on recursive side and prover side, so we can put it as is by now.

tonyfloatersu (approved):
generally lgtm, was thinking in a hyperplonk style of merging 2 PCS opening into 1, but this one also makes sense to me (actually they are the same in mechanism).

randomly leaving some thoughts in comments, feel free to ignore

PR #197, opened 2025-02-11
Zz/dev on chain verifier

zhenfeizhang:


transcript/src/transcript.rs
```diff
@@ -380,7 +382,8 @@ where
     fn generate_circuit_field_element(&mut self) -> <ChallengeF as ExtensionField>::BaseField {
         if !self.data_pool.is_empty() {
             self.hash_state = self.hasher.hash_to_state(&self.data_pool);
-            self.data_pool.clear();
+            // reset the data_pool from the output of last hash
+            self.data_pool = self.hash_state.clone();
```
tonyfloatersu: I see what is going wrong here, but my proposal is on a distinct direction, that is:
hash to state should be input with self.hash state appending data pool, then data pool should still be cleared.

I use data pool intentionally for new appended elements, but adding hash state back to the data pool does not seem right to me, as it makes code think there are something new in the data pool, and not use the preexisting elements in current hash state, which incurs a few more hashes.

PR #200, opened 2025-02-14
PCS: Hyrax support

tonyfloatersu:


arith/src/bn254.rs
```diff
@@ -163,6 +166,44 @@ impl FieldSerde for Fr {
     }
 }
 
+impl FieldSerde for G1Affine {
+    const SERIALIZED_SIZE: usize = 32;
```
tonyfloatersu, line 170: This is taken from #191, I am a bit unsure if my G1/G2-Affine serialized size is correct...

but for a larger perspective - do we actually need the serialize size anymore?  Is this constant used anywhere else in the codebase, or we can just eliminate it altogether?

poly_commit/src/hyrax/hyrax_impl.rs
```diff
@@ -0,0 +1,132 @@
...
+    randomness: &mut Vec<C::Scalar>,
+) -> HyraxCommitment<C>
+where
+    C::Scalar: ExtensionField + PrimeField,
+    C::ScalarExt: ExtensionField + PrimeField,
+{
+    let vars = mle_poly.num_vars();
+    let pedersen_vars = (vars + 1) / 2;
+    let pedersen_len = 1usize << pedersen_vars;
+    assert_eq!(pedersen_len, params.bases.len());
```
tonyfloatersu, line 72: Here is another question - I am reordering the whole multilinear poly into a square-alike matrix, should we make it more like Orion way, that the matrix is not square-alike but rather rather wide (1024 bits in cols)?

The motivation of the question is out of perf concern, that will serial multiexp be performing faster if the input MSM is sufficient long (which stands for Orion way there), or actually not much difference out there?

poly_commit/src/hyrax/pedersen.rs
```diff
@@ -0,0 +1,91 @@
...
+    params: &PedersenParams<C>,
+    coeffs: &[C::Scalar],
+    r: C::Scalar,
+) -> C
+where
+    C::Scalar: PrimeField,
+{
+    let mut what = C::default().to_curve();
+    msm::multiexp_serial(coeffs, &params.bases[..coeffs.len()], &mut what);
+    msm::multiexp_serial(&[r], &[params.h], &mut what);
```
tonyfloatersu: my pedersen commitment considers hiding property - binding directly holds from vanilla pedersen.

I mean, simulatability is almost free, do we want it?

poly_commit/src/orion/simd_field_agg_impl.rs
```diff
@@ -83,12 +83,9 @@ where
...
 
-    if !orion_mt_verify(
-        vk,
-        &query_indices,
-        &proof.query_openings[..query_num],
-        &roots[0],
-    ) {
+    if izip!(proof.query_openings.chunks(query_num), &roots)
+        .any(|(paths, r)| !orion_mt_verify(vk, &query_indices, paths, r))
+    {
```
tonyfloatersu, line 88: this is a fixing to prior impl - supposing only root of MPI machine verifies, then eventually it should check all the MT openings, I only took care of the first party, i.e., root's MT opening, should just verify all of them

poly_commit/src/hyrax/inner_prod_argument.rs
```diff
@@ -0,0 +1,213 @@
...
+            .collect()
+    };
+    let (com_r, rs_mask) = pedersen_commit_ro_absorb(&r_vec);
+
+    let (com_zero, zero_mask) = pedersen_commit_ro_absorb(&[]);
+
+    let ip_ry: C::Scalar = izip!(&r_vec, y_vec).map(|(r, y)| *r * *y).sum();
+    let (com_ip_ry, ip_mask) = pedersen_commit_ro_absorb(&[ip_ry]);
+
+    let mu: C::Scalar = transcript.generate_challenge_field_element();
```
tonyfloatersu: the thing surprises me is that - in jolt impl there is actually no IPA, they sent over the folded row of multilinear poly, and verify runs eval check and runs pedersen commitment to see if it matches the linear combination of prior commitments, which is similar to the "raw" pcs in a sense.

supposing we are not looking for zero knowledge property here, I suppose we can just ditch the IPA here and just return the top var fixed multilinear poly, which save a bit on randomness and perf?

tonyfloatersu (commented):
self running a round of review for your reference, a bit of questions leaving around the code and certain concerns around give-and-takes on designs and motivations

poly_commit/src/hyrax/inner_prod_argument.rs
tonyfloatersu: yea removing whole IPA and sigma protocol completely in https://github.com/PolyhedraZK/Expander/pull/200/commits/57d3a1559bc78c148fb18b9fdd6986fb768ba67d for no simulatability

poly_commit/src/hyrax/pedersen.rs
tonyfloatersu: rm simulatability in https://github.com/PolyhedraZK/Expander/pull/200/commits/57d3a1559bc78c148fb18b9fdd6986fb768ba67d

arith/src/bn254.rs
zhenfeizhang: 32 is correct. 

https://github.com/PolyhedraZK/Expander/blob/main/arith/mersenne31/src/m31.rs#L42

this is used by the deserializer. 

I am actually not sure about the following:

```
 fn deserialize_from<R: Read>(reader: R) -> FieldSerdeResult<Self> {
        let bytes: Vec<u8> = Vec::deserialize_from(reader)?;
        let mut encoding = <Self as GroupEncoding>::Repr::default();
        encoding.as_mut().copy_from_slice(bytes.as_ref());
        match G1Affine::from_bytes(&encoding).into_option() {
            Some(a) => Ok(a),
            None => Err(FieldSerdeError::DeserializeError),
        }
    }
```

I have not used this type of deserialization before. it seems that you read all bytes from R. maybe it is sufficient to check the length of `encoding` and `bytes` here?

arith/src/bn254.rs
```diff
@@ -163,6 +166,44 @@ impl FieldSerde for Fr {
...
+
+    fn serialize_into<W: Write>(&self, writer: W) -> FieldSerdeResult<()> {
+        let bytes = self.to_bytes().as_ref().to_vec();
+        bytes.serialize_into(writer)
+    }
+
+    fn deserialize_from<R: Read>(reader: R) -> FieldSerdeResult<Self> {
+        let bytes: Vec<u8> = Vec::deserialize_from(reader)?;
```
zhenfeizhang, line 178: see previous comments

arith/src/bn254.rs
```diff
@@ -163,6 +166,44 @@ impl FieldSerde for Fr {
...
+impl FieldSerde for G2Affine {
+    const SERIALIZED_SIZE: usize = 64;
+
+    fn serialize_into<W: Write>(&self, writer: W) -> FieldSerdeResult<()> {
+        let bytes = self.to_bytes().as_ref().to_vec();
+        bytes.serialize_into(writer)
+    }
+
+    fn deserialize_from<R: Read>(reader: R) -> FieldSerdeResult<Self> {
+        let bytes: Vec<u8> = Vec::deserialize_from(reader)?;
```
zhenfeizhang, line 201: see previous comments

poly_commit/src/hyrax/hyrax_impl.rs
zhenfeizhang: well this is tricky....

let's say you have n rows and each row has m elements and you want to do an MSM for this m elements. you end up with a cost of 

`c = n * m / log(m) `

so basically you have a fixed `n * m` and you try to minimize for `c`. the smallest `c` is obviously when `n = 1` and therefore no hyrax at all. 

or I should say that for the best of prover performance we should use things like bullet proof rather than hyrax.

For verifier, iirc, your complexity is `O(n) + O(m)` so it makes sense to build a square matrix rather then the Orion one?

poly_commit/src/hyrax/hyrax_impl.rs
```diff
@@ -0,0 +1,117 @@
...
+        .collect();
+
+    HyraxCommitment(commitments)
+}
+
+pub(crate) fn hyrax_open<C>(
+    params: &PedersenParams<C>,
+    mle_poly: &impl MultilinearExtension<C::Scalar>,
+    eval_point: &[C::Scalar],
+) -> (C::Scalar, Vec<C::Scalar>)
```
zhenfeizhang: nits: I think I'd prefer a struct `proof` or `opening` that wraps `Vec<C::Scalar>` like you did here
```
#[derive(Clone, Debug, Default)]
pub struct HyraxCommitment<C: CurveAffine + FieldSerde>(pub Vec<C>);
```


also add a comment that the first `C::Scalar` is the eval of the polynomial

poly_commit/src/hyrax/pedersen.rs
```diff
@@ -0,0 +1,65 @@
...
+    length: usize,
+    mut rng: impl rand::RngCore,
+) -> PedersenParams<C>
+where
+    C::Scalar: PrimeField,
+{
+    let bases: Vec<C> = (0..length)
+        .map(|_| {
+            let scalar = C::Scalar::random(&mut rng);
+            (C::generator() * scalar).to_affine()
```
zhenfeizhang: collect all the projective elements first and then use `batch_affine`

tonyfloatersu: sure - see https://github.com/PolyhedraZK/Expander/pull/200/commits/48b0df2dafa16dfac526d28057ad0b0827ca8444

poly_commit/src/hyrax/hyrax_impl.rs
tonyfloatersu: see https://github.com/PolyhedraZK/Expander/pull/200/commits/48b0df2dafa16dfac526d28057ad0b0827ca8444

arith/src/bn254.rs
tonyfloatersu: fair - see https://github.com/PolyhedraZK/Expander/pull/200/commits/48b0df2dafa16dfac526d28057ad0b0827ca8444

I checked against the bytes read from the reader, should match the defined `SERIALIZED_SIZE`.

poly_commit/src/hyrax/hyrax_impl.rs
tonyfloatersu: > so basically you have a fixed n * m and you try to minimize for c. the smallest c is obviously when n = 1 and therefore no hyrax at all.

If MSM behaves as "the more you MSM, the more you save" as there is an inverse log m term, then surely deteriorate to dimension 1 is the way to go for prover... but not what we want

> or I should say that for the best of prover performance we should use things like bullet proof rather than hyrax.

I am unaware of bulletproof, haven't read through the paper, only have an impression that it is related to dory and other IPA with log (?) proof size, that somehow uses randomness to reduce the sqrt proof size down to log?

But how is this related to the prover perf?  I think the bottleneck is still MSM here?

> For verifier, iirc, your complexity is O(n) + O(m) so it makes sense to build a square matrix rather then the Orion one?

Correct, either way you gotta do O(m + n) group operations, i.e., linear combine commitments then pedersen commit the opening - you are screwed either way (actually the complexity is `O(n / log(n) + m / log(m))` if you use MSM in pedersen commitment and commitment linear combination)

zhenfeizhang (approved):
LGTM!

PR #203, opened 2025-02-20
Fix Poseidon benchmark

devfans:
Looks like a misconfig

### DIFF
--- a/gkr/src/main.rs
+++ b/gkr/src/main.rs
@@ -18,7 +18,7 @@ use gkr::{
         KECCAK_BN254_CIRCUIT, KECCAK_BN254_WITNESS, KECCAK_GF2_CIRCUIT, KECCAK_GF2_WITNESS,
         KECCAK_M31_CIRCUIT, KECCAK_M31_WITNESS, POSEIDON_M31_CIRCUIT, POSEIDON_M31_WITNESS,
     },
-    BN254ConfigSha2Raw, GF2ExtConfigSha2Orion, M31ExtConfigSha2Orion, Prover,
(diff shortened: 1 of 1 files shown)


tonyfloatersu:
Thank you for your attention, but we consider this being not a misconfig - we want to integrate the poseidon benchmarking against GKR argument system with a polynomial commitment scheme other than "raw", which one considered backup option for PCS.

For further explanation, the "poseidon" in the config you changed stands for the Fiat-Shamir hash being instantiated by poseidon hash over M31, but it does not stand for the benchmark content.

devfans:
Thanks for your quick response. However without the change, the benchmark gives me errors below:
```

===============================
benchmarking poseidon with GKR^2 over m31ext3
field:          m31ext3
#threads:       8
#bench repeats: 4
hash scheme:    poseidon
===============================
Warning: padding witnesses, expected 16, got 1
Circuit loaded!
We are now calculating average throughput, please wait for 5 seconds
thread '<unnamed>' panicked at /Users/stefan/git/Expander/poly_commit/src/orion/simd_field_impl.rs:73:9:
assertion `left == right` failed
  left: 15
 right: 11 

```
With main branch code. Any idea?

tonyfloatersu:
i see what's going wrong, fault on our end, my apologies - this thing is wired to the `gkr_square`, which was implemented a while ago without other optimations, i.e., simd/paralellization... so there are a few rounds of sumcheck missing, which lead to the `15 == 11` assertion failure - namely simdx16 stands for 4 rounds/vars missing.

We are now running a fix at this point, but for now you can revert back the relevant changes in #192 to the prior config, for your local benchmarking purpose.

PR #204, opened 2025-02-21
Fix gkr square for simd mpi

zhiyong1997:


niconiconi:
we should have a git hook for auto lint @zhiyong1997 https://git-scm.com/book/ms/v2/Customizing-Git-Git-Hooks

PR #207, opened 2025-02-25
Gkr2

zhiyong1997:
Major changes credit to @enpsi20481


gkr/src/verifier.rs
```diff
@@ -40,6 +43,10 @@ fn verify_sumcheck_step<C: GKRFieldConfig, T: Transcript<C::ChallengeField>>(
         *claimed_sum = GKRVerifierHelper::degree_2_eval(&ps, r, sp);
     } else if degree == 3 {
         *claimed_sum = GKRVerifierHelper::degree_3_eval(&ps, r, sp);
+    } else if degree == 6 {
```
niconiconi: hardcoded degree may be an future issue

gkr/src/verifier.rs
```diff
@@ -283,7 +290,7 @@ impl<Cfg: GKRConfig> Verifier<Cfg> {
         // this function will iteratively hash the commitment, and append the final hash output
         // to the transcript. this introduces a decent circuit depth for the FS transform.
         //
-        // note that this function is almost identical to grind, expect that grind uses a
+        // note that this function is almost identical to grind, except that grind uses a
         // fixed hasher, where as this function uses the transcript hasher
         let pcs_verified = transcript.append_commitment_and_check_digest(&buffer, &mut cursor);
```
niconiconi, line 306: fix the FS attack by enable grinding by default

sumcheck/src/prover_helper/sumcheck_gkr_square.rs
```diff
@@ -1,62 +1,136 @@
...
+        if self.mpi_config.is_root() {
+            global_vals.try_into().unwrap()
+        } else {
+            [C::ChallengeField::ZERO; D]
+        }
     }
 
     #[inline]
-    pub(crate) fn receive_challenge(&mut self, var_idx: usize, r: C::ChallengeField) {
+    pub(crate) fn poly_evals_at_simd(&self, var_idx: usize) -> [C::ChallengeField; D] {
```
niconiconi, line 90: maybe you need to merge these two function, they are duplicated

sumcheck/src/scratch_pad.rs
```diff
@@ -143,6 +146,45 @@ impl<C: GKRFieldConfig> VerifierScratchPad<C> {
             deg3_lag_denoms_inv[i] = denominator.inv().unwrap();
         }
 
+        let deg6_eval_at = if C::FIELD_TYPE == FieldType::GF2 {
+            // TODO: Does this correctly define Lagrange poly for GF2?
```
niconiconi: no
please add a assertion fail for GF2 here

niconiconi (changes requested):

gkr/src/verifier.rs
zhiyong1997: By default, the transcript will compute `PCS_DIGEST_LOOP` (currently = 1000) hashes after receiving the initial commitment. See line 126 and line 308 in `transcript/src/transcript.rs`. However, when the `recursion` feature is turned on, these hashes are skipped for now because it may significantly increase the size of the recursion circuit, and we've yet thought of a solution.

sumcheck/src/prover_helper/sumcheck_gkr_square.rs
zhiyong1997: Are you referring to the the `mpi combine` part in `poly_evals_at_x` and `poly_evals_at_simd`? i.e. line 87 and line 107.

sumcheck/src/scratch_pad.rs
zhiyong1997: Safe guard at the beginning of `gkr_square_prove` and `gkr_square_verify`.

gkr/src/verifier.rs
zhiyong1997: Slightly improved this hardcoded problem in the new commit.

niconiconi (approved):
LGTM

PR #127, opened 2024-10-17
Recursive Verifier

zhiyong1997:

PR #211, opened 2025-02-27
Change the required number of witnesses in MPI

zhiyong1997:


gkr/src/exec.rs
```diff
@@ -23,16 +23,6 @@ async fn main() {
...
-    if let ExpanderExecSubCommand::Verify {
-        witness_file: _,
-        input_proof_file: _,
-        mpi_size,
-    } = &expander_exec_args.subcommands
-    {
-        assert_eq!(mpi_config.world_size, 1);
-        mpi_config.world_size = *mpi_size as i32;
-    }
-
```
tonyfloatersu, line 35: somehow this concerns me a bit... this was introduced in #127 gkr/srs/exec.rs, and I modified to this implementation on commandline revamp, then this PR is removing it...

I wonder why in the first place we are introducing mpi into verifier tho, and if possible, is it fine we just remove MPI totally from verifier side?

gkr/src/executor.rs
```diff
@@ -186,17 +187,15 @@ pub async fn run_command<'a, Cfg: GKRConfig>(command: &ExpanderExecArgs, config:
             input_proof_file,
             mpi_size,
         } => {
+            assert!(
+                config.mpi_config.world_size() == 1,
+                "Do not run verifier with mpiexec."
+            );
+            config.mpi_config.world_size = mpi_size as i32;
```
tonyfloatersu, line 194: ok - I was confused in the first place, that the world size rewinding moved to this place, I see...

circuit/src/layered/circuit.rs
```diff
@@ -148,49 +149,97 @@ impl<C: GKRFieldConfig> Circuit<C> {
...
+        self.load_witness_bytes(&file_bytes, mpi_config, false, false);
     }
 
-    pub fn load_witness_bytes(&mut self, file_bytes: &[u8], allow_padding: bool) {
+    pub fn load_witness_bytes(
+        &mut self,
+        file_bytes: &[u8],
+        mpi_config: &MPIConfig,
+        is_prover: bool,
+        allow_padding_for_testing: bool, // TODO: Consider remove this
```
tonyfloatersu, line 178: the line on 158 is a bit hurting to me - hard to infer what 2 true in arguments meaning...

This actually starts to make me wonder what is the landscape and where we are at right now - I knew that we previously discussed that separating public input and witness, maybe this will help somewhat with `is_prover`, as verifier can load just the public input rather than reading the whole witness...

I guess the allow padding serves the same purpose of align statement-witness pairs up to pack size, now also with a factor of MPI size - but still pad with the first statement-witness pair, among all MPI parties?

tonyfloatersu (commented):
im done with my first pass, certain concerns going to be addressed in offline comm

circuit/src/layered/circuit.rs
tonyfloatersu, line 178: another round of fact check: if I understood correctly, this essentially means following things:
- Each MPI party read a **full** set of statement/witness pairs, that in fact, each party just need a segment of these pairs
- The witness file should be assumed to be looking like - SIMD x MPI statement/witness pairs are concatenated, that eventually if we want to pad against full *desired num of witness*, we copy the first statement/witness pair to align with that number (which suppose to be a po2).

gkr/src/main.rs
```diff
@@ -132,9 +132,10 @@ fn run_benchmark<Cfg: GKRConfig>(args: &Args, config: Config<Cfg>) {
     };
 
     match args.scheme.as_str() {
-        "keccak" => circuit_template.load_witness_file(witness_path),
+        "keccak" => circuit_template.prover_load_witness_file(witness_path, &config.mpi_config),
         "poseidon" => match Cfg::FieldConfig::FIELD_TYPE {
-            FieldType::M31 => circuit_template.load_non_simd_witness_file(witness_path),
+            FieldType::M31 => circuit_template
+                .load_witness_allow_padding_testing_only(witness_path, &config.mpi_config),
```
tonyfloatersu: I mean - this does not look like the case in main mpi or gkr correctness, I suppose the real difference lies in the padding option being true or false - yet if I recall correctly - keccak case should be fully padded, so the padding option being true does not matter here, so I advocate changing to just use load witness allow padding API here for simplicity and alignment?

tonyfloatersu (commented):
final round of review, consider this as a fact check on my understanding to the real deal of the code changes.

circuit/src/layered/circuit.rs
```diff
@@ -148,49 +149,97 @@ impl<C: GKRFieldConfig> Circuit<C> {
...
                     "Warning: padding witnesses, expected {}, got {}",
-                    C::get_field_pack_size(),
-                    witness.num_witnesses
+                    desired_number_of_witnesses, witness.num_witnesses
                 );
+                let padding_vec = witness.values[0..total_size].to_vec();
+                for _ in witness.num_witnesses..desired_number_of_witnesses {
+                    witness.values.extend_from_slice(&padding_vec);
+                }
+                witness.num_witnesses = desired_number_of_witnesses;
```
tonyfloatersu, line 215: on a second thought - this copying and extending does not have to happen on all the MPI node - supposing the MPI node has high rank yet insufficient number of witnesses, this should happen, while on the other case, this does not have to happen... but this is jut an observation that does not have to be implemented?

circuit/src/layered/circuit.rs
zhiyong1997: That's correct! I'll optimize the witness loading with shared memory in the future.

gkr/src/main.rs
zhiyong1997: Good point!

circuit/src/layered/circuit.rs
zhiyong1997: That's correct. I thought the witness loading efficiency might not be important for testing, but I'll optimize this a little bit in the shared-memory version.

tonyfloatersu: awesome!

tonyfloatersu (approved):
my questions resolved, ship it.

PR #214, opened 2025-03-04
unify serdes for circuit

zhenfeizhang:
cross ref https://github.com/PolyhedraZK/ExpanderCompilerCollection/pull/115

- use a unified `ExpSerdes` trait for all serialization APIs
- removed `FieldForECC`, merged API into `Field`
- implemented `PartialEq` for field equality check (and checked edge cases for 0 == MODULUS)


zhenfeizhang:
~tests are failing as the circuit serde is changed~

arith/gf2/src/gf2x8.rs
```diff
@@ -42,6 +44,8 @@ impl Field for GF2x8 {
 
     const INV_2: Self = GF2x8 { v: 0 }; // should not be used
 
+    const MODULUS: U256 = U256::ZERO; // should not be used
```
tonyfloatersu: maybe use `unimplemented!()`?

circuit/src/serde.rs
```diff
@@ -1,63 +1,57 @@
...
+};
 
 use super::{Allocation, CoefType, Gate, RecursiveCircuit, Segment, Witness};
 use crate::{GateAdd, GateConst, GateMul, SegmentId};
 
-#[derive(Debug, Error)]
-pub enum CircuitError {
-    #[error("field serde error: {0:?}")]
-    FieldSerdeError(#[from] FieldSerdeError),
+impl<C: GKRFieldConfig, const INPUT_NUM: usize> ExpSerde for Gate<C, INPUT_NUM> {
```
tonyfloatersu, line 12: minor: I think this is somewhat colliding with #220?

arith/gf2_128/src/gf2_ext128x8/avx256.rs
```diff
@@ -104,6 +102,8 @@ impl Field for AVX256GF2_128x8 {
 
     const FIELD_SIZE: usize = 128;
 
+    const MODULUS: U256 = U256::ZERO; // should not be used
```
tonyfloatersu: same, probably just `unimplement` this

poly_commit/src/raw.rs
```diff
@@ -2,26 +2,25 @@
...
+impl<F: Field> ExpSerde for RawCommitment<F> {
     const SERIALIZED_SIZE: usize = unimplemented!();
 
-    fn serialize_into<W: std::io::Write>(&self, mut writer: W) -> FieldSerdeResult<()> {
-        let u256_embedded = U256::from(self.evals.len() as u64);
-        let fr_embedded = BN254Fr::from_u256(u256_embedded);
-        fr_embedded.serialize_into(&mut writer)?;
+    fn serialize_into<W: std::io::Write>(&self, mut writer: W) -> SerdeResult<()> {
+        let len = self.evals.len();
+        writer.write_all(len.to_le_bytes().as_ref())?;
```
tonyfloatersu: If I recall correctly, this dumb (de)serialization of length to 256bit width has a reason behind, and I was part of it:
- This change was introduced back in #163, where raw PCS serialization length into u64 causes problem on recursion side, where recursion is reading in a unit of u256 for curve
- We moved on to lift length into u256, to make recursion back to life, probably we want to keep the serialization as it was, namely serialize to u256 rather than u64?

arith/src/field.rs
```diff
@@ -34,7 +34,11 @@ pub trait Field:
     + for<'a> AddAssign<&'a Self>
     + for<'a> SubAssign<&'a Self>
     + for<'a> MulAssign<&'a Self>
-    + FieldSerde
+    + ExpSerde
+    + Hash
```
tonyfloatersu, line 38: just curious: what is the motivation of introducing `Hash` trait in `Field`?

arith/mersenne31/src/m31.rs
```diff
@@ -90,14 +114,67 @@ impl Field for M31 {
...
         (rng.next_u32() & 1).into()
     }
 
+    #[inline(always)]
+    fn to_u256(&self) -> U256 {
+        U256([self.v as u128, 0])
+    }
+
+    #[inline(always)]
+    fn from_u256(value: U256) -> Self {
```
tonyfloatersu, line 133: also curious: is rewriting this method for better perf, as prev modulus is between u256, while now you are doing m31 modulus as 8 chunks of u32 divisions?

arith/gf2/src/gf2.rs
```diff
@@ -115,19 +128,19 @@ impl Field for GF2 {
...
+        U256([self.v as u128, 0])
+    }
 
-    fn from_u256(x: ethnum::U256) -> Self {
+    #[inline(always)]
+    fn from_u256(value: U256) -> Self {
+        // unsafe -- we assume the value is in the field
         GF2 {
-            v: (x.as_u32() & 1) as u8,
+            v: (value.0[0] % 2) as u8,
```
tonyfloatersu, line 141: a bit unsure of the ideas behind - sometimes the from method auto corrects inputs, while some other cases the from method bails if the input is not in range [0, modulus - 1] - do we have a rule behind these method implementation?

arith/src/field.rs
siq1: In ECC, there are some `HashMap`s with field elements as keys. Previously, there was a trait `FieldForECC`, but now it's merged into `Field`.

arith/mersenne31/src/m31.rs
siq1: This method is called in very few scenarios, performance doesn't matter much.

arith/gf2/src/gf2.rs
siq1: I think it should auto correct inputs.

arith/src/field.rs
tonyfloatersu: I see, that makes sense, as it should be supported as a hashed key

arith/gf2/src/gf2x8.rs
zhenfeizhang: fixed f9403a4656a044439233f0e99b972fad67bbc5c8

poly_commit/src/raw.rs
zhenfeizhang: ok, reverted ea70049b90cb522c699e29d82aa0352fe1f22e75

zhenfeizhang: > where recursion is reading in a unit of u256 for curve

I think it may be better solution to fix recursion part to read 8 bytes instead of 32...

tonyfloatersu (approved):
let's push this forward and see how it goes down the road

PR #220, opened 2025-03-07
Shared Memory for Circuit

zhiyong1997:
This PR adds the shared memory feature, allowing multiple mpi processes to share the same circuit without allocating additional memory. The program logic should be correct, but the interface may need to be refactored, and the extensive use of raw pointers may need to be minimized. Any suggestions are appreciated.


circuit/src/ecc_circuit.rs
```diff
@@ -102,25 +102,25 @@ impl<C: GKRFieldConfig> RecursiveCircuit<C> {
...
                     }
                     for gate in &leaf_seg.gate_consts {
-                        let mut gate = gate.clone();
+                        let mut gate = *gate;
                         gate.o_id += alloc.o_offset;
                         ret_layer.const_.push(gate);
                     }
                     for gate in &leaf_seg.gate_uni {
-                        let mut gate = gate.clone();
+                        let mut gate = *gate;
```
tonyfloatersu, line 124: seems to be `Copy`?

circuit/src/layered/gates.rs
```diff
@@ -21,3 +21,7 @@ pub type GateMul<C> = Gate<C, 2>;
 pub type GateAdd<C> = Gate<C, 1>;
 pub type GateUni<C> = Gate<C, 1>;
 pub type GateConst<C> = Gate<C, 0>;
+
+// If I Simply add derive(Copy) to the Gate struct, the compiler does not seem to recognize it
+// for the type aliases. Explicitly state it here.
+impl<C: GKRFieldConfig, const INPUT_NUM: usize> Copy for Gate<C, INPUT_NUM> {}
```
tonyfloatersu, line 27: yea seems you `Copy` the `Gate` here

circuit/src/layered/serde.rs
tonyfloatersu, line 1: serialization/deserialization, boilerplate code.  I guess you want to hold off a bit as #214 is out there... just notifying

tonyfloatersu: ngl the structured impl for CoefType -> Gate -> CircuitLayer -> Circuit is giving me a good sense of what circuit data structure should look like, that's rather helpful.

config/gkr_field_config/src/traits.rs
```diff
@@ -14,7 +14,7 @@ pub trait GKRFieldConfig: Default + Debug + Clone + Send + Sync + 'static {
     const FIELD_TYPE: FieldType;
 
     /// Field type for the circuit, e.g., M31
-    type CircuitField: Field + FieldForECC + Send;
+    type CircuitField: Clone + Field + FieldForECC + Send;
```
tonyfloatersu: wait... isn't Field already clone and copy?

circuit/src/layered/circuit.rs
```diff
@@ -144,11 +145,33 @@ impl<C: GKRFieldConfig> Clone for Circuit<C> {
 unsafe impl<C> Send for Circuit<C> where C: GKRFieldConfig {}
 
 impl<C: GKRFieldConfig> Circuit<C> {
-    pub fn load_circuit<Cfg: GKRConfig<FieldConfig = C>>(filename: &str) -> Self {
+    // Load a circuit from a file and flatten it
+    // Used for single-thread prover or verifier
+    pub fn load_circuit_independent<Cfg: GKRConfig<FieldConfig = C>>(filename: &str) -> Self {
```
tonyfloatersu: minor: suggest renaming to `load circuit single core/process`? maybe we should stick to certain naming convention when it comes to MPI multi-procses or single process vanilla settings

circuit/src/layered/circuit.rs
```diff
@@ -144,11 +145,33 @@ impl<C: GKRFieldConfig> Clone for Circuit<C> {
...
+    // called before the end of the program
+    pub fn load_circuit_shared<Cfg: GKRConfig<FieldConfig = C>>(
+        filename: &str,
+        mpi_config: &MPIConfig,
+    ) -> (Self, *mut ompi_win_t) {
+        let circuit = if mpi_config.is_root() {
+            Some(Self::load_circuit_independent::<Cfg>(filename))
+        } else {
+            None
+        };
```
tonyfloatersu: starts to make me wonder about the motivation of this PR: is this the case that, on a single machine but with multicore, prior circuit loading means - each core/process loads one copy of circuit each, and thus in RAM there are multiple redundant shares of circuits, share number equals to core number.

With the introduction of such change, we can bring the RAM consumption down to constant, i.e., 1 circuit loading across the multiple core via shared memory?

circuit/src/layered/circuit.rs
```diff
@@ -144,11 +145,33 @@ impl<C: GKRFieldConfig> Clone for Circuit<C> {
...
+
+        let (mut circuit, window) = mpi_config.consume_obj_and_create_shared(circuit);
+        circuit.pre_process_gkr::<Cfg>();
```
tonyfloatersu, line 189: wait... but isn't you already ran `pre_process_gkr` in `flatten`, which is in load circuit independent? why we want to run again here?

circuit/src/layered/circuit.rs
tonyfloatersu: on a second thought, I suppose this method will only be used by verifier, as eventually we are deprecating MPI config from verifier, so only root verifier reads the circuit...

so a minor suggestion: do you want to align the main.rs and main_mpi.rs implementations such that they will only call the load circuit shared such that the implementation looks aligned?

circuit/src/layered/circuit.rs
```diff
@@ -361,14 +384,20 @@ impl<C: GKRFieldConfig> Circuit<C> {
         self.rnd_coefs_identified = true;
     }
 
-    pub fn fill_rnd_coefs<T: Transcript<C::ChallengeField>>(&mut self, transcript: &mut T) {
+    pub fn fill_rnd_coefs<T: Transcript<C::ChallengeField>>(
+        &mut self,
+        transcript: &mut T,
+        mpi_config: &MPIConfig,
+    ) {
```
tonyfloatersu: I suppose this method will also be used in verifier, that is ran by only the root process...

gkr/src/executor.rs
```diff
@@ -216,7 +222,11 @@ pub async fn run_command<'a, Cfg: GKRConfig>(command: &ExpanderExecArgs, mut con
                 .collect::<Vec<u8>>()
                 .try_into()
                 .unwrap();
-            let circuit = Circuit::<Cfg::FieldConfig>::load_circuit::<Cfg>(&command.circuit_file);
+            let circuit = Circuit::<Cfg::FieldConfig>::load_circuit_shared::<Cfg>(
```
tonyfloatersu: maybe just write as `let (circuit, _) = ...` ?

tonyfloatersu: or just directly use load circuit independent?

tonyfloatersu (commented):
This is a first pass of the code, haven't really touched the real deal in MPI config for shared memory, but this leaves me an impression of shared memory being helpful to single machine multi core circuit loading RAM usage case.  Am I understanding the motivation correctly here?

circuit/tests/circuit_serde.rs
```diff
@@ -0,0 +1,69 @@
...
+    FiatShamirHashType::SHA256,
+    PolynomialCommitmentType::Raw
+);
+
+declare_gkr_config!(
+    pub BN254ConfigSha2Raw,
+    FieldType::BN254,
+    FiatShamirHashType::SHA256,
+    PolynomialCommitmentType::Raw
+);
```
tonyfloatersu, line 34: the code redundancy makes me start wondering if we want to move these commonly used configs from gkr into config?

circuit/src/layered/shared_mem.rs
```diff
@@ -0,0 +1,96 @@
+use super::circuit::{Circuit, CircuitLayer, StructureInfo};
+use super::gates::{GateAdd, GateConst, GateMul, GateUni};
+use gkr_field_config::GKRFieldConfig;
+use mpi_config::shared_mem::SharedMemory;
+
+impl<C: GKRFieldConfig> SharedMemory for CircuitLayer<C> {
+    fn bytes_size(&self) -> usize {
+        8 + 8
```
tonyfloatersu, line 8: I suppose this is `input/output var num` here? please consider explicitly writing so, rather than deriving magic number here

circuit/src/layered/serde.rs
```diff
@@ -0,0 +1,127 @@
...
+    fn deserialize_from<R: Read>(mut reader: R) -> FieldSerdeResult<Self> {
+        let layers = Vec::<CircuitLayer<C>>::deserialize_from(&mut reader)?;
+        Ok(Circuit {
+            layers,
+
+            public_input: vec![],
+            expected_num_output_zeros: 0,
+
+            rnd_coefs_identified: false,
+            rnd_coefs: vec![],
```
tonyfloatersu: I suppose you can just `..Default::default()` here?

circuit/src/layered/shared_mem.rs
```diff
@@ -0,0 +1,96 @@
...
+}
+
+impl<C: GKRFieldConfig> SharedMemory for Circuit<C> {
+    fn bytes_size(&self) -> usize {
+        8 + self
+            .layers
+            .iter()
+            .map(|layer| layer.bytes_size())
+            .sum::<usize>()
+            + 8
```
tonyfloatersu: yet again there are magic number here... also I wonder if you can just implement `SharedMemory` for `Vec<T>` where `T: SharedMemory` such that the byte size here don't have to do sum here?

circuit/src/layered/serde.rs
zhiyong1997: The serialization/deserialization here is only for debugging purpose, we can adapt to the implementation in that PR.

config/gkr_field_config/src/traits.rs
zhiyong1997: My bad, was trying to to debug why the compiler didn't recognize the `COPY` trait.

circuit/src/layered/circuit.rs
zhiyong1997: Good suggestion, I'll remove the use of `mpi_config` for the verifier, and make some adjustment to the function name.

circuit/src/layered/circuit.rs
zhiyong1997: Exactly! I'm targeting at saving one circuit in the memory for all the mpi processes.

circuit/src/layered/circuit.rs
zhiyong1997: Yes, this is a little weird.. I was thinking about backward compatibility in ECC, but on second thought, let's make it more organized, and change ECC accordingly.

gkr/src/prover/linear_gkr.rs
```diff
@@ -103,14 +103,17 @@ impl<Cfg: GKRConfig> Prover<Cfg> {
...
         grind::<Cfg>(&mut transcript, &self.config);
 
-        c.fill_rnd_coefs(&mut transcript);
+        if self.config.mpi_config.is_root() {
+            c.fill_rnd_coefs(&mut transcript);
+        }
+        self.config.mpi_config.barrier();
         c.evaluate();
 
+        transcript_root_broadcast(&mut transcript, &self.config.mpi_config);
```
tonyfloatersu, line 137: I am a bit unsure about changing the execution sequence of fill randomness into circuit, then transcript broadcast... I guess if you merge back the recursion fix, then something would complain on recursive verifier side...

I guess this code runs after prover load circuit, so would fill randomness affect the copy that is read by other parties tho?

tonyfloatersu (commented):
I downloaded the code to my local end and still playing with it, the question I need help with is the transcript broadcast thing here...

Other than that, I am still reading the MPI manual to see if things should work accordingly

circuit/src/layered/shared_mem.rs
zhiyong1997: Ideally, I would like to implement `SharedMemory` for all types that implement `Copy`, and then all `Vec<T>` where `T` implements `SharedMemory`. However, the compiler would complain about potential conflict implementation saying that some upstream may implement `Copy` for `Vec<_>` in the future. Do you have any idea on how to solve this problem?

gkr/src/prover/linear_gkr.rs
zhiyong1997: Sorry for the confusion. Let me explain this a little bit. Since everyone has a shared memory now, the update on the root rank would also take effect on other ranks. What `fill_rnd_coefs` does now is the root rank fill in the randomness, while others wait. However, after the execution, the transcript states on all ranks are different, and thus a sync up is required. 

For the recursion, I think we can change it accordingly without too much effort.

tonyfloatersu: I see it now - so the prev logic is syncing the randomness, then each party agrees on randomness, then the randomness filled in the circuit is the same; while now, only root party fill the randomness, then move on to sync the randomness with other party for Fiat-Shamir later on.  I see the corresponding changes in recursion, so that looks reasonable to me.

tonyfloatersu (commented):
seems good to me, waiting on a merge back from main to resolve the conflict then we approve and merge

tonyfloatersu (approved):
Let's roll it out first!

PR #222, opened 2025-03-11
update CLI; Fix CI's MPI issue

zhenfeizhang:
This PR does three things:

- improved UI for CLI -- previously circuit file has to be supplied before sub-command which is confusing for users
- fix CI issues -- cargo lock pinned an earlier version of openmpi 
- auto cancel CI tasks when there is a newer commit

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -5,6 +5,11 @@ env:
   RUST_BACKTRACE: 1
   ACTIONS_RUNNER_DEBUG: true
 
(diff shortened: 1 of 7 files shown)


tonyfloatersu (commented):
seems fine with me - mostly just putting circuit file from general arg to sub command?

tonyfloatersu (approved):

zhenfeizhang:
> seems fine with me - mostly just putting circuit file from general arg to sub command?

yes.

PR #223, opened 2025-03-11
PCS: Reduce excessive matrix transposes in Orion

tonyfloatersu:
We are attempting to reduce the number of matrix transposes in Orion down to 1, while the transpose should happen after expander code encoding, before MT commitment - eventually the transpose should happen across multiple MPI cores.


poly_commit/src/orion/utils.rs
```diff
@@ -227,39 +210,26 @@ pub(crate) fn transpose_in_place<F: Field>(mat: &mut [F], scratch: &mut [F], row
 }
 
 #[inline(always)]
-pub(crate) fn transpose_and_pack<F, SimdF>(evaluations: &mut [F], row_num: usize) -> Vec<SimdF>
+pub(crate) fn pack_from_base<F, PackF>(evaluations: &[F]) -> Vec<PackF>
```
zhiyong1997, line 214: Is this a switch from packing multiple rows to packing adjacent elements in each row?

tonyfloatersu: exactly - just pack neighboring elements (the variables are neighboring), such that we want to avoid excessive transposes for packing across rows

poly_commit/src/orion/base_field_tests.rs
```diff
@@ -20,23 +20,34 @@ where
...
-        .flat_map(|msg| orion_srs.code_instance.encode(msg).unwrap())
-        .collect();
+    let mut interleaved_codeword = {
+        let mut rows_of_codeword: Vec<_> = poly
+            .coeffs
+            .chunks(msg_size * ComPackF::PACK_SIZE)
+            .flat_map(|chunk| -> Vec<_> {
+                let mut chunk_local = chunk.to_vec();
+                let mut scratch = chunk_local.clone();
+                transpose_in_place(&mut chunk_local, &mut scratch, msg_size);
```
zhiyong1997: I feel like it's a little too costly to allocate an additional `O(N)` memory to transpose a matrix `in place`.

tonyfloatersu: oh, good observation.  this is just a proof of concept in testing, to see how vanilla transpose work like, eventually matching the real commitment.

but I sort of agree, that I should at least move the scratch out, s.t., the O(N) memory complexity can be lifted out of the loop.

zhiyong1997 (approved):
After the discussion, I'm now convinced that the new ordering of coefficients is correct and sound. Good job!

tonyfloatersu:
block a bit on #214 remerge

PR #234, opened 2025-03-22
Minor: SIMD field method horizontal sum, and a weird application

tonyfloatersu:
tldr: M31x16 horizontal sum can skip a bunch of mod reductions, then Orion M31 opening gets 50% faster.

### DIFF
--- a/arith/mersenne31/src/m31x16/m31_avx256.rs
+++ b/arith/mersenne31/src/m31x16/m31_avx256.rs
@@ -284,15 +284,47 @@ impl SimdField for AVXM31 {
 
     const PACK_SIZE: usize = M31_PACK_SIZE;
 
+    #[inline(always)]
     fn pack(base_vec: &[Self::Scalar]) -> Self {
         assert_eq!(base_vec.len(), M31_PACK_SIZE);
         let ret: [Self::Scalar; M31_PACK_SIZE] = base_vec.try_into().unwrap();
         unsafe { transmute(ret) }
     }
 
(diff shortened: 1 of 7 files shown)


arith/mersenne31/src/m31x16/m31_avx256.rs
```diff
@@ -284,16 +284,48 @@ impl SimdField for AVXM31 {
...
     fn unpack(&self) -> Vec<Self::Scalar> {
         let ret = unsafe { transmute::<[__m256i; 2], [Self::Scalar; M31_PACK_SIZE]>(self.v) };
         ret.to_vec()
     }
+
+    #[inline(always)]
+    fn horizontal_sum(&self) -> Self::Scalar {
+        let mut buffer = 0u64;
+        let ret = unsafe { transmute::<[__m256i; 2], [Self::Scalar; M31_PACK_SIZE]>(self.v) };
+
```
zhenfeizhang: let buffer = ret.iter().map(|r|r.v as u64).sum();

zhenfeizhang (approved):

arith/mersenne31/src/m31x16/m31_avx256.rs
tonyfloatersu: sure, see https://github.com/PolyhedraZK/Expander/pull/234/commits/084fca8fe01317ba4287fc5bf47614a01850db5d

I was trying to do loop unrolling manually, but it seems as good I guess?

PR #225, opened 2025-03-13
Poly: FFT and univariate polynomial

tonyfloatersu:

PR #235, opened 2025-03-23
impl goldilocks field

zhenfeizhang:
- [x] require FFTField from https://github.com/PolyhedraZK/Expander/pull/225
- [x] cross ref https://github.com/PolyhedraZK/ExpanderCompilerCollection/pull/126


serdes/src/serdes.rs
```diff
@@ -203,3 +203,26 @@ impl<K: ExpSerde + Eq + Hash, V: ExpSerde> ExpSerde for HashMap<K, V> {
         Ok(map)
     }
 }
+
+impl<T: ExpSerde> ExpSerde for Option<T> {
```
tonyfloatersu: maybe remerge again with main, I think this has already been in main?

arith/goldilocks/src/goldilocksx8/goldilocks_avx512.rs
```diff
@@ -0,0 +1,436 @@
...
+
+#[derive(Debug, Clone, Copy)]
+pub struct AVXGoldilocks {
+    // each __m512i element contains 8 Goldilocks elements
+    pub v: __m512i,
+}
+
+impl AVXGoldilocks {
+    #[inline(always)]
+    pub fn pack_full(x: Goldilocks) -> Self {
```
tonyfloatersu, line 41: a bit confused here: seems `From<Goldilocks>` is redundant with `pack_full` method?

arith/goldilocks/src/goldilocks.rs
```diff
@@ -0,0 +1,579 @@
...
+impl From<u64> for Goldilocks {
+    #[inline(always)]
+    fn from(x: u64) -> Self {
+        Goldilocks {
+            v: if x < GOLDILOCKS_MOD {
+                x
+            } else {
+                x - GOLDILOCKS_MOD
+            },
+        }
```
tonyfloatersu: minor: just use `mod_reduce_u64`?

arith/goldilocks/src/goldilocks_ext.rs
```diff
@@ -0,0 +1,376 @@
...
+        }
+        let z0square = z0 * z0;
+
+        let mut res = [Goldilocks::ZERO; 2];
+
+        res[0] = arr[0] * z0;
+        res[1] = arr[1] * z0square;
+
+        Self { v: res }
+    }
```
tonyfloatersu, line 375: I have a sense that this section can be simplified, but it smells like plonky code.  The `count` thing here reaching below line 356 can either be 0 or 1, while 0 case is self, 1 case is `(-a + bX)` for input being `(a + bX)`.

I think a easier way for inv here follows: observing input `(a + bX)` has `(a - bX)` yielding product back to base field, denoted as `d` - then the multiplicative inverse is `d^{-1} (a - bX)`, which should be straightforward.

Another question that I am a bit dumb here: seems Frobenius automorphis is `X \mapsto X^p`, but why it looks to me here that `(a + bX)^p` should look like `a + bX` rather than `-a + bX` here ...?  If you have some more cycles please let me know where I did my calc wrong...

tonyfloatersu (approved):

tonyfloatersu (approved):
I am ok with the current implementation, seems going through E2E test in `gkr_correctness`, and a few auxiliary tests for base/extension/SIMD settings, still some minor comments in code reading with some confusions

arith/goldilocks/src/goldilocksx8/goldilocks_avx512.rs
zhenfeizhang: they are the same...

PR #236, opened 2025-03-24
PCS: Orion MPI transpose and MPI MT opening

tonyfloatersu:
This PR was finalizing the purpose of "single matrix transpose during Orion commit" on MPI scale, with detailed changes as follows:

1. As previously suggested - invoking Orion commit with MPI world size > 1, a single MPI matrix transpose is invoked, detailed implementation see `mpi_utils.rs`:

   * For matrix transpose on local single threaded settings, we replaced in house prototyped transpose mechanism by introducing `transpose` crate as one of the dependencies - perf seems to be slightly improving.
   * For matrix transpose on multiple core, i.e., MPI world size > 1, we introduced MPI all to all transpose in prior #242, and finalized the matrix transpose in `mpi_utils.rs` with one single all to all transpose and a few additional sub transposes.
   * Eventually, the query complexity of MPI invoked PCS opening is equivalent to the one from a single process.

2. Code simplification by removing base field impls with unified into SIMD impl, together with a global verifier method.

3. Orion input robustness: there are 2 considerations w.r.t. MT query opening size:

   * On a rather small polynomial input, the polynomial might be even smaller than a single MT query opening
   * As the polynomial size scales up, the opening size should be at least $O(\log \ell)$ when polynomial size is $O(\ell)$.

   We introduced a new method `orion_eval_shape` in `utils.rs` to decide the minimum admissible polynomial size, and global MT query opening size.

   NOTE that this method is only invoked once during PCS setup with polynomial size as inputs, then it is only decided once during setup before the proving task starts.

   If the polynomial is too small - the GKR system need to scale up the polynomial by pad to sufficient boolean hypercube size - see collateral changes in `gkr` crate.

   This `orion_eval_shape` method supersedes the prior `minimum_num_vars` method, that was invoked as an oracle at any moment, while right now is (and should be) only invoked once during PCS setup, and gives sufficient information in CRS and GKR system.

   Another collateral change lies in both PCS traits (expander PCS and the one in `poly_commit`), deprecating `minimum_num_vars` with updated `gen_crs_for_testing` returning - returning with suggested polynomial number of variables tied to the CRS generated.

4. MT memory optimizations - removed a few methods invoking internal copying, directly use ownership moving and memory transmute saves memory bandwidth.

5. Goldilocksx8 settings in Orion introduced, tested, and benchmarked, updated Goldilocksx8 SIMD horizontal sum.

6. Updated MT hashes to Keccak.


gkr_engine/src/poly_commit/definition.rs
```diff
@@ -19,25 +19,26 @@ pub trait ExpanderPCS<F: FieldEngine> {
 
     const PCS_TYPE: PolynomialCommitmentType;
 
-    type Params: Clone + Debug + Default + Send + Sync + 'static;
+    type Params: Clone + Debug + Default + Send + Sync + Into<usize> + 'static;
```
zhiyong1997: I think it's a little confusing: what does `Into<usize>` mean for `Params`? I think we can either introduce a `Params` trait or simply add more comments here.

gkr_engine/src/poly_commit/definition.rs
```diff
@@ -19,25 +19,26 @@ pub trait ExpanderPCS<F: FieldEngine> {
...
+    /// that the PCS might need to accept a polynomial of extended length.
+    ///
+    /// NOTE(HS) the calibrated number of variables refers to the local SIMD variables
+    /// rather than the base field elements.
     fn gen_srs_for_testing(
         params: &Self::Params,
         mpi_engine: &impl MPIEngine,
         rng: impl RngCore,
-    ) -> Self::SRS;
+    ) -> (Self::SRS, usize);
```
zhiyong1997, line 52: I feel like the returned `usize` is equal to `params.into()`, is this correct?

zhiyong1997 (approved):
I've went through the overall changes, but yet to take a look at every single detail in `Orion`.

gkr_engine/src/poly_commit/definition.rs
tonyfloatersu: yea that's right, I think your prior observation about param Into usize is also correct - that I want to cast params into number of variables for PCS on invocation.

Maybe I should make a trait for params down the road with a explicit method casting itself into a number of variable?

tonyfloatersu: yea nailed it in https://github.com/PolyhedraZK/Expander/pull/236/commits/24f4b60dbec91844dc375eb71fb9ac86a81c07dc

gkr_engine/src/poly_commit/definition.rs
tonyfloatersu: yea, see https://github.com/PolyhedraZK/Expander/pull/236/commits/24f4b60dbec91844dc375eb71fb9ac86a81c07dc that I wrapped up the PCS params trait with an explicit num vars method

PR #238, opened 2025-03-25
refactoring binaries

zhenfeizhang:


tonyfloatersu (approved):
seems ok with me - I think the script modifications only take effect on local end not on the CI end?

zhenfeizhang:
> seems ok with me - I think the script modifications only take effect on local end not on the CI end?

yes

PR #240, opened 2025-03-27
Huge refactoring: unify all trait definitions

zhenfeizhang:
Huge refactoring: unify all trait definitions

- [x] merge #238 first

PR #242, opened 2025-03-28
Building blocks for Orion MPI matrix transpose PR

tonyfloatersu:
This PR contains following things separated from #236: 
- A method in Eq poly, on given an index representing small endian `\iv \in {0, 1}^\ell`, compute `\eq(\rv, \iv)`
  Here the single query need $O(\ell)$ time, for computing `\eq(\rv, *)` over whole boolean hypercube, don't do it.

- Avoiding a few copying in MT building through direct memory transmuting.

- Two MPI related methods:
  - All-To-All transpose, that supposing each party holds a row of matrix (assuming the length mod party number is 0), run MPI all to all method to exchange sections of their own row to other parties.
  - MPI gather various length array, the motivation here is gathering MT openings from other parties - invokes 2 rounds of MPI communication, 1 for letting root party know the number of bytes from each party, 1 for gather varcount MPI invocation.  We cannot assume what the elements sent between parties, so `ExpSerde` (de)serialization needed.

A few heads up:
- #240 might need to move the MPI methods below `mpi_engine`
- a bit unsure if shared memory #220 can help with the gather various len MPI here, leaving a note here


config/mpi_config/src/lib.rs
```diff
@@ -272,6 +275,95 @@ impl MPIConfig {
...
+            .for_each(|(row_chunk, recv_chunk)| {
+                row_chunk[copy_starts..copy_ends].copy_from_slice(recv_chunk);
+            });
+
+            copy_starts += send_buffer_size_per_world;
+        });
+    }
+
+    #[inline(always)]
+    pub fn gather_varlen_vec<F: ExpSerde>(&self, elems: &Vec<F>, global_elems: &mut Vec<Vec<F>>) {
```
zhiyong1997, line 333: I think we can avoid explicit calling of `(de)serialization` if `F` implements `Copy`, but I'm fine with current implementation.

config/mpi_config/src/lib.rs
```diff
@@ -272,6 +275,95 @@ impl MPIConfig {
...
+    /// supposing the current party holds a row in a matrix with row number being MPI parties.
+    #[inline(always)]
+    pub fn all_to_all_transpose<F: Sized>(&self, row: &mut [F]) {
+        assert_eq!(row.len() % self.world_size(), 0);
+
+        // NOTE(HS) MPI has some upper limit for send buffer size, pre declare here and use later
+        const SEND_BUFFER_MAX: usize = 1 << 22;
+
+        let row_as_u8_len = size_of_val(row);
+        let row_u8s: &mut [u8] =
```
zhiyong1997, line 288: This looks better than `vec_to_u8_bytes` and `leak`. I'll switch to this in some other PR.

zhiyong1997 (approved):
Looks good to me.

config/mpi_config/src/lib.rs
tonyfloatersu: valid concern, but this is for gathering MT opening, the total communication complexity is around 5(ish) MB.

will add more comments about motivation in following PR

PR #243, opened 2025-03-29
Zz/impl babybear

zhenfeizhang:
- [x] dependency: #240


arith/gf2_128/src/gf2_ext128x8/avx512.rs
```diff
@@ -360,8 +354,8 @@ impl Debug for AVX512GF2_128x8 {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         let mut data = [0u8; 128];
         unsafe {
-            _mm512_storeu_si512(data.as_mut_ptr() as *mut i32, self.data[0]);
-            _mm512_storeu_si512((data.as_mut_ptr() as *mut i32).offset(16), self.data[1]);
+            _mm512_storeu_si512(data.as_mut_ptr() as *mut __m512i, self.data[0]);
+            _mm512_storeu_si512((data.as_mut_ptr() as *mut __m512i).offset(16), self.data[1]);
```
tonyfloatersu: offset should be 1?

tonyfloatersu (commented):
overall looks reasonable, there is one catch in avx512 gf2_128 fmt, think the pointer should point to wrong place?

arith/gf2_128/src/gf2_ext128x8/avx512.rs
zhenfeizhang: nice catch!

tonyfloatersu (approved):

PR #249, opened 2025-04-03
Fix some field operations

siq1:
Fixes:
1. Goldilocks `try_inverse_u64` should check `x == GOLDILOCKS_MOD`
2. Goldilocks `from_u256` should handle `value >= GOLDILOCKS_MOD`
3. In Goldilocks and M31, some functions uses the raw value `v`, including `to_u256`, `Hash`, `PartialOrd`, `Ord`. But the inner `v` might be `>= MOD`. We should reduce it first.

These bugs make ECC Goldilocks tests failed. Bug 3 might affect M31, but it didn't. Maybe it's because the probability is too low for M31.

### DIFF
--- a/arith/goldilocks/src/goldilocks.rs
+++ b/arith/goldilocks/src/goldilocks.rs
@@ -25,7 +25,7 @@ pub(crate) fn mod_reduce_u64(x: u64) -> u64 {
     }
 }
 
-#[derive(Debug, Clone, Copy, Default, PartialOrd, Ord)]
+#[derive(Debug, Clone, Copy, Default)]
 pub struct Goldilocks {
     pub v: u64,
 }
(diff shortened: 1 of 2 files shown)


siq1:
Performance impact:
1. not sure
2. In M31, `from_u256` has a custom implementation, maybe we can implement one for Goldilocks too?
3. `to_u256`, `Hash`, `PartialOrd`, `Ord` are only used in ECC, and are used in very few cases.

zhenfeizhang:
before this PR 911e0c26bb3b0c081e167a08f5639887c53c9035

```
cargo run --release --bin gkr -- -f m31ext3
0-bench: throughput: 410 hashes/s

cargo run --release --bin gkr -- -f m31ext3 -c poseidon
0-bench: throughput: 157820 hashes/s
```

after this PR 752c762d00b0c5bcbc2dfe6c85324ef8a3bea224

```
cargo run --release --bin gkr -- -f m31ext3
0-bench: throughput: 403 hashes/s

cargo run --release --bin gkr -- -f m31ext3 -c poseidon
0-bench: throughput: 168188 hashes/s
```

seems within the noise range

zhenfeizhang (approved):

PR #252, opened 2025-04-08
fix goldilocks

zhenfeizhang:
added `FFTField` tests for goldilocks and found out that `Goldilocksx8` was implemented incorrectly.

### DIFF
--- a/arith/goldilocks/src/goldilocksx8/goldilocks_avx512.rs
+++ b/arith/goldilocks/src/goldilocksx8/goldilocks_avx512.rs
@@ -322,7 +322,7 @@ impl FFTField for AVXGoldilocks {
 #[inline]
 unsafe fn mod_reduce_epi64(x: __m512i) -> __m512i {
     // Compare each element with modulus
-    let mask = _mm512_cmpgt_epu64_mask(x, PACKED_GOLDILOCKS_MOD);
+    let mask = _mm512_cmpge_epu64_mask(x, PACKED_GOLDILOCKS_MOD);
     // If element > modulus, subtract modulus
     _mm512_mask_sub_epi64(x, mask, x, PACKED_GOLDILOCKS_MOD)
 }
--- a/arith/goldilocks/src/goldilocksx8/goldilocks_avx512.rs
+++ b/arith/goldilocks/src/goldilocksx8/goldilocks_avx512.rs
@@ -36,17 +36,6 @@ pub struct AVXGoldilocks {
     pub v: __m512i,
 }
 
-impl AVXGoldilocks {
-    #[inline(always)]
-    pub fn pack_full(x: Goldilocks) -> Self {
-        unsafe {
-            Self {
-                v: _mm512_set1_epi64(x.v as i64),
-            }
-        }
-    }
-}
-
 field_common!(AVXGoldilocks);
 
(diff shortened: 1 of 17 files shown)


arith/goldilocks/src/goldilocksx8/goldilocks_avx512.rs
```diff
@@ -322,7 +322,7 @@ impl FFTField for AVXGoldilocks {
 #[inline]
 unsafe fn mod_reduce_epi64(x: __m512i) -> __m512i {
     // Compare each element with modulus
-    let mask = _mm512_cmpgt_epu64_mask(x, PACKED_GOLDILOCKS_MOD);
+    let mask = _mm512_cmpge_epu64_mask(x, PACKED_GOLDILOCKS_MOD);
     // If element > modulus, subtract modulus
     _mm512_mask_sub_epi64(x, mask, x, PACKED_GOLDILOCKS_MOD)
```
tonyfloatersu: maybe `>= modulus` as `cmpgt` changed to `cmpge`?

arith/goldilocks/src/goldilocksx8/goldilocks_neon.rs
tonyfloatersu, line 1: the separate implementation of fft field leaves me an impression that maybe `pack_full` should be a `SIMDField` trait method, can be done directly through memory transmute from a slice of base field elements, slice length is the `SIMDField::PACK_SIZE`...

tonyfloatersu (approved):
seems fine - goldilocks mod reduction is basically apply strict reduction between `[0, p - 1]` rather than `[0,p]`, which eventually caused trouble

arith/goldilocks/src/goldilocksx8/goldilocks_neon.rs
zhenfeizhang: i think it is a great idea to move pack_full into SIMD trait.

>  can be done directly through memory transmute from a slice of base field elements,

i am not sure this works... it propagates a elements into multiple slots.

``` rust
impl AVXGoldilocks {
    #[inline(always)]
    pub fn pack_full(x: Goldilocks) -> Self {
        unsafe {
            Self {
                v: [
                    _mm256_set1_epi64x(x.v as i64),
                    _mm256_set1_epi64x(x.v as i64),
                ],
            }
        }
    }
}
```

Are you talking about the pack function?

``` rust
    #[inline(always)]
    fn pack(base_vec: &[Self::Scalar]) -> Self {
        assert_eq!(base_vec.len(), Self::PACK_SIZE);
        let base_vec_array: [Self::Scalar; 8] = base_vec.try_into().unwrap();
        unsafe { transmute(base_vec_array) }
    }
```

tonyfloatersu: > > can be done directly through memory transmute from a slice of base field elements,
>
> i am not sure this works... it propagates a elements into multiple slots.

never mind - I was thinking directly using something like `transmute::<[F; SimdF::PACK_SIZE], SimdF>` to achieve the goal of only one implementation for `SimdF::pack_full`, yet a counter example I have in mind is the GF2 case, just directly contradict with the implementation, so the implementation so far I am good with.

PR #257, opened 2025-04-13
Field: Mersenne31 ext6

tonyfloatersu:


arith/mersenne31/src/m31_ext6.rs
```diff
@@ -0,0 +1,327 @@
...
+        self.v[0].as_u32_unchecked()
+    }
+
+    #[inline(always)]
+    fn from_uniform_bytes(bytes: &[u8; 32]) -> Self {
+        let v1 = mod_reduce_u32(u32::from_be_bytes(bytes[0..4].try_into().unwrap()));
+        let v2 = mod_reduce_u32(u32::from_be_bytes(bytes[4..8].try_into().unwrap()));
+        let v3 = mod_reduce_u32(u32::from_be_bytes(bytes[8..12].try_into().unwrap()));
+
+        // TODO(HS) need some inputs - do we need extra mod reduce u32 like m31ext3 impl?
```
zhenfeizhang: there is a small chance that for uniform u32 inputs, a single mod reduce will produce 2^31 or 2^31-1 so i think we should take two mod reduction here

arith/mersenne31/src/m31_ext6.rs
```diff
@@ -0,0 +1,327 @@
...
+        }
+    }
+
+    fn random_bool(mut rng: impl RngCore) -> Self {
+        M31Ext6 {
+            v: [M31Ext3::random_bool(&mut rng), M31Ext3::ZERO],
+        }
+    }
+
+    fn inv(&self) -> Option<Self> {
```
zhenfeizhang, line 95: i don't really understand this algorithm...

tonyfloatersu: oh, I was doing something tricky, gotta explain here:

The M31Ext3Ext2, i.e., M31Ext6, is $F_{p^6} = F_{p^3} [Y] / (Y^2 +  2)$, and we let $a + b Y \in F_{p^6}$ and $c + d Y \in F_{p^6}$ are mutually inverse each other.

Then $(a + bY) (c + dY) \equiv ac + (bc + ad) Y + bd Y^2 \equiv (ac - 2bd) + (bc + ad) Y \mod Y^2 + 2$.  Now we want $ac - 2bd = 1$ and $bc + ad = 0$.

We can make the second condition true first - suppose we have $a + bY \in F_{p^6}$, $-a + bY$ satisfies, then we normalize back the scalar field back to 1.

$(a + bY) (-a + bY) \equiv (-a^2 - 2b^2) \mod Y^2 + 2$, then we do the `normalize`  variable on line 100.

Eventually, $n = (-a^2 - 2b^2)^{-1} \in F_{p^3}$, then result is $n \times (-a + bY) \in F_{p^6}$.

tonyfloatersu: the whole point is to avoid the huge exp from Fermat little theorem...

zhenfeizhang (approved):

tonyfloatersu:
will remove the TODO comment before merge back.

arith/mersenne31/src/m31_ext6.rs
tonyfloatersu: im a bit unsure so I am unresolving this one - seems `mod_reduce_u32` reduces inputs into range `[0, p]`, but rather we want something like `[0, p - 1]`, an additional call to `mod_reduce_u32` won't help

asking for suggestions here - do we want to fix the behavior that m31 from random bytes to strict `[0,p-1]` range? otherwise, we can just remove the additional call to mod reduce u32, as that won't help...

tonyfloatersu: see https://github.com/PolyhedraZK/Expander/pull/257/commits/ac7889a83a66fe927461f14d5b27ece06449c91a per offline discussion.

PR #261, opened 2025-04-15
Zq/overload operator

chonpsk:
overload operator
generalize transcript


arith/mersenne31/src/tests.rs
```diff
@@ -117,9 +117,18 @@ fn test_poseidon_m31_fiat_shamir_hash() {
     let perm = PoseidonFiatShamirHasher::<M31x16>::new();
 
     {
-        let state_elems: [M31; M31x16::RATE] = [M31::from(114514); M31x16::RATE];
-        let actual_output = perm.hash_to_state(&state_elems);
-        let expected_output = vec![
+        // let state_elems: [M31; M31x16::RATE] = [M31::from(114514); M31x16::RATE];
```
zhenfeizhang, line 120: remove

arith/mersenne31/src/poseidon.rs
```diff
@@ -25,4 +25,8 @@ impl PoseidonStateTrait for M31x16 {
     fn full_round_sbox(&mut self) {
         *self = self.exp(Self::SBOX_POW as u128);
     }
+
+    fn to_u8(&self, output: &mut [u8]) {
```
zhenfeizhang, line 29: 1. is there a reason that `to_u8` takes a mutable output instead of return a output. 
2. or perhaps a trait bound `PoseidonStateTrait: Into<&[u8]>`?
3. `to_u8_slices` may also be a better name if we don't use the trait bound

arith/src/bn254.rs
```diff
@@ -75,9 +75,9 @@ impl Field for Fr {
 
     #[inline(always)]
     // TODO: better implementation
-    fn from_uniform_bytes(bytes: &[u8; 32]) -> Self {
+    fn from_uniform_bytes(bytes: &[u8]) -> Self {
         <Fr as FromUniformBytes<64>>::from_uniform_bytes(
-            &[bytes.as_slice(), [0u8; 32].as_slice()]
+            &[&bytes[..32], [0u8; 32].as_slice()]
```
zhenfeizhang, line 80: add `assert!(bytes.len()>=32)`

arith/src/field.rs
```diff
@@ -90,7 +91,21 @@ pub trait Field:
...
+    fn to_bytes(&self, target: &mut [u8]) {
+        let mut res = vec![];
+        self.serialize_into(&mut res).unwrap();
+        target.copy_from_slice(&res);
+    }
+
+    /// convert from bytes
+    #[inline(always)]
+    fn set_in_bytes(&mut self, source: &[u8]) {
+        *self = Self::deserialize_from(Cursor::new(source)).unwrap();
```
zhenfeizhang, line 107: `*self = Self::deserialize_from(source).unwrap();`

gkr_engine/src/transcript/definition.rs
```diff
@@ -9,79 +9,104 @@ use super::Proof;
...
         &mut self,
         commitment_bytes: &[u8],
         proof_reader: &mut R,
     ) -> bool;
 
     /// Append a field element to the transcript.
-    fn append_field_element(&mut self, f: &F);
+    #[inline]
+    fn append_field_element<F: Field>(&mut self, f: &F) {
+        let mut buf = vec![];
```
zhenfeizhang, line 35: not sure about this... previously for field hashes

``` rust
  if !self.proof_locked {
            let mut buffer = vec![];
            f.serialize_into(&mut buffer).unwrap();
            self.proof.bytes.extend_from_slice(&buffer);
        }
        self.data_pool.extend_from_slice(&f.to_limbs());
```
and this is likely what caused the recursion to fail

chonpsk: I think the result is the same as bytes hashes and the implement of bytes hashes is more elegant so I unified them. Since the len of the sample slice was modified, the transcript result will differ. If it tests the recurision comparing to go version and go version uses the old transcript it makes sense that failed.
Or try to test recursion with bytes and field separately to comfirm it.

arith/mersenne31/src/poseidon.rs
chonpsk: just to reduce memory operations. If using a trait or return a output directly aligns rust's conventions better, I will modify it.
3. agree

gkr_engine/src/transcript/definition.rs
tonyfloatersu: my bad for being late for the requested convo - was caught up prototyping FRI thing

iirc the motivation of separating `byte hash transcript` from `field hash transcript` was that - the internal reserved "sponge-alike" structure in both transcripts holds different types - for former, the basic unit is a `u8`, while for latter, the basic unit is a `F: Field` - which explains why transcript carries a type argument in type declaration.

(there are a few other details in impls w.r.t., `field hash transcript` holding a hash state for drawing unused samples from one hash, but that is relatively irrelevant)

I am unsure where the convo is heading, yet my observation is that the change on `field hash transcript` side is rather large, that I can't quite see how these two behaviors will be the same, yet coerce field elements on recursion side to be bytes, then go on forward makes it a bit painful to rewrite the recursion transcript, I guess.

Issue #218, opened 2025-03-05
Decoupling MPI from verifier


tonyfloatersu:
During a few rounds of code reviews and code readings, we observed that MPI config also resides in current verifier construction, yet verifier is not supposed/assumed to run over MPI, later on the parallelized verifier is also not designed to utilize the MPI's parallelism, so we decide to draft an issue on decoupling MPI from verifier.

The current status of using MPI in verifier is that - we want to know the "world size" in one proving session, namely this "world size" help with deciding how many statement/witness pairs are involved in the proving.  The only information and utilization of a MPI config instance needed in the verifier is the "world size".

In this sense, verifier should only maintain a usize, rather than a whole ref to MPI config instance.

A few related things in my mind to jot down:
- Removing MPI config from PCS verification, in a sense the current PCS opening should give same opening among all parties, though only root opening is used in verification.
- If possible, separating statement and witness in the witness file, so verifier will not have to read the whole chunk (which in turns clobber the verifier's asymptotic perf).  Yet this is a bit out of the hand of only prover.

PR #269, opened 2025-04-23
MPI-free verifier example

weikengchen:
This PR includes a version of Expander GKR module (which would be needed to verify a proof) without dependency on MPI.

Experiments only. Do not merge.


zhenfeizhang:
cross ref #218

PR #271, opened 2025-04-24
generate proof binaries; refactoring executables

zhenfeizhang:


gkr/src/utils.rs
tonyfloatersu, line 1: I think you can also move this `utils.rs` to be under `bin`, as I think it only serve the purpose of keeping track of where the circuit (witness) files should go?

tonyfloatersu (commented):
looks generally fine, just one comment

gkr/src/utils.rs
zhenfeizhang: `gkr_correctness` test will need the file names defined in `utils.rs`

tonyfloatersu (approved):

PR #280, opened 2025-05-06
implement exp_serde derive

zhenfeizhang:
break changes:

turns out we don't need `const SERIALIZED_SIZE` in `ExpSerde` and remove this makes the derivation a lot more generic.


serdes_derive/src/lib.rs
```diff
@@ -0,0 +1,147 @@
...
+            }
+        }
+        Data::Enum(data) => {
+            let variants = data.variants;
+            let _variant_count = variants.len();
+            let mut serialize_arms = Vec::new();
+            let mut deserialize_arms = Vec::new();
+            for (idx, variant) in variants.iter().enumerate() {
+                let vname = &variant.ident;
+                let variant_index = idx + 1;
```
zhenfeizhang: enum idx starts at 1 to match ECC

cc @siq1

poly_commit/src/raw.rs
```diff
@@ -17,9 +17,11 @@ pub struct RawCommitment<F: Field> {
 }
 
 impl<F: Field> ExpSerde for RawCommitment<F> {
-    const SERIALIZED_SIZE: usize = unimplemented!();
-
     fn serialize_into<W: std::io::Write>(&self, mut writer: W) -> SerdeResult<()> {
+        // ideally we want to remove this and use Derive(ExpSerde) method to determine the length
```
zhiyong1997, line 21: For recursion, we now require all the bytes fed into the transcript to be interpretable as a sequence of field elements so that field-based hasher like MIMC5 or Poseidon will have the same behavior as it is in Expander. This problem can be resolved, but may need some more engineering work.

zhiyong1997 (approved):

PR #281, opened 2025-05-08
[Minor] Correct the size of eq_evals_first/second_half

zhiyong1997:


zhenfeizhang (approved):

zhenfeizhang:
this is a performance issue and will not affect the correctness right? as if the size is smaller than the actual size, the mutable buffer will simply increase the size.

PR #282, opened 2025-05-12
Fix a corner case in build_eq_x_r_helper

zhiyong1997:
The updated behaviour now alignes with `build_eq_x_r_with_buf`.


zhenfeizhang (approved):
Nice catch!

PR #286, opened 2025-05-20
[feat] defer kzg pairing check

zhenfeizhang:
# design rational

In KZG verification, there is a pairing check. 
When we do batch verification of many PCSs, we can defer all the pairing checks and do them together. This will reduce the #final exponentiation from O(k) to 1.

# what is changed

Added APIs 

- partial check for PCS (minus final pairing check)
- an accumulator that stores pairing inputs that are to be checked later


zhenfeizhang:
end up using #290

PR #288, opened 2025-05-21
Replacing static value in MPI Config

zhiyong1997:
In our current implementation, a global static value is used to store the mpi communicator. Although this works in Expander, it limits the power of the mpi communication toolkit, prohibiting its usage when there are multiple worlds and communicators. This PR targets at removing the global static value, and replacing it with a local value of proper lifetime.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2038,6 +2038,7 @@ version = "0.1.0"
 dependencies = [
  "ethnum",
  "halo2curves",
+ "rand",
  "serdes_derive",
  "thiserror",
 ]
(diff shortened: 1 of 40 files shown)


zhiyong1997:
/gemini Thanks for the valuable feedback. For MPI Finalize, this is a intended removal though. Previously, we used a global static value to store the `universe` object, which never reaches the end of its lifetime, and that's why we need to explicitly do MPI Finalize. However, in the updated version, the universe object usually has a non-static lifetime, and it will automatically finalize the mpi environment at the end of its lifetime. I removed the MPI Finalize because this may cause confusion, and lead to double finalize. Do you agree with this, or do you have any suggestions?

zhiyong1997:
/gemini The `unsafe impl send` for MPI is removed now. Do you have any other concerns?

zhiyong1997:
/gemini The design of the executor is indeed a little wierd, but the change is beyond the scope of this PR. I temporarily add some comments to avoid confusion of multi-process and single-process.

zhenfeizhang (approved):

PR #305, opened 2025-07-24
Allow variable-length polys in SumOfProduct Poly and PCS Claims Merging

zhiyong1997:
In sumcheck for the `SumOfProduct` Poly, we required all the polynomials to have the same length. In some cases where there were smaller polynomials, we padded them to max length, and then commit, merge, and open. This worked, but were not very efficient. This PR made the following changes:

1. Allow variable-length polys in `SumOfProduct` poly without padding overhead.
2. Adapt the sumcheck protocol to the new poly, increasing efficiency.
3. In merging multiple PCS claims and challenges, remove the padding overhead, the overall cost is linear to the total size now.

### DIFF
--- a/arith/polynomials/src/sum_of_products.rs
+++ b/arith/polynomials/src/sum_of_products.rs
@@ -1,6 +1,6 @@
 use arith::Field;
 
-use crate::{MultiLinearPoly, MultilinearExtension};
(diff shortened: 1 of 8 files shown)


zhenfeizhang (approved):
LGTM

PR #318, opened 2026-01-27
Fix verifier not verifying

jsgold-1:
### Summary
___
This PR fixes a verifier bug introduced by the recent change, which binds public inputs to the Fiat–Shamir transcript. This change caused existing tests and proofs/verifications to no longer pass.

With this PR, the verifier now deserializes public inputs from the proof, checks them against the provided public_input, and hashes the verified values into the transcript (as before) before challenge generation. This restores the expected transcript flow and allows codebase tests to run successfully.

This change is limited to input handling and is intended purely as a compatibility/correctness fix.


pauldelucia (approved):
This fixes the issue for me.

hczphn (approved):

PR #323, opened 2026-02-11
Remove MPI dependency

HudsonGraeme:
## Summary
- Eliminate OpenMPI crate and all MPI-based distributed proving infrastructure
- Simplify `MPIConfig` to a plain struct (no lifetime parameter, no MPI universe/communicator)
- All `MPIEngine` trait methods now operate as single-process no-ops (gather copies, broadcast is identity, barrier is no-op)
- Remove `gkr-mpi` binary target and `main_mpi.rs`
- `prover_load_circuit` returns `Self` instead of `(Self, MPI_Win)`
- Remove openmpi from CI workflows

## Test plan
- [x] `cargo check` passes
- [ ] CI runs clean on this branch
- [ ] Downstream ECC builds against this branch (see inference-labs-inc/ecc rm-mpi branch)


HudsonGraeme:
Opened on wrong repo, moving to inference-labs-inc/Expander

PR #325, opened 2026-02-22
Optimize SRS serialization: parallel uncompressed format, file caching, buffered I/O

hczphn:
## Summary

- **Parallel uncompressed SRS serialization** (`structs_kzg.rs`): Replace derived `ExpSerde` on `CoefFormUniKZGSRS` with custom implementation using uncompressed G1 point format + rayon parallel conversion. Deserialization uses `from_uncompressed_unchecked` to skip expensive square root computation. SRS loading goes from minutes to seconds.
- **SRS file caching** (`poly_commit/src/utils.rs`): Add `expander_pcs_init_with_srs_path()` — generates SRS once and saves to disk; subsequent runs load from file. `expander_pcs_init_testing_only` now defaults to `/tmp/kzg_srs_{n}.bin` cache.
- **64MB buffered I/O** (`definition.rs`): Wrap SRS file reads/writes in `BufReader`/`BufWriter` with 64MB capacity to reduce syscall overhead.
- **Vec pre-allocation** (`serdes.rs`): Use `Vec::with_capacity(len)` instead of `Vec::default()` in generic Vec deserialization to avoid repeated reallocation.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/gkr_engine/src/poly_commit/definition.rs
+++ b/gkr_engine/src/poly_commit/definition.rs
@@ -6,6 +6,9 @@ use std::{fmt::Debug, str::FromStr};
 
 use crate::{ExpErrors, ExpanderSingleVarChallenge, FieldEngine, MPIEngine, Transcript};
 
+/// Buffer capacity for SRS file I/O (64 MB).
+const SRS_IO_BUFFER_CAPACITY: usize = 64 * 1024 * 1024;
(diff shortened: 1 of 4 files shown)


hczphn:
@gemini review

